PR #52, opened 2015-11-26
Rating buttons now display when dark theme is on Closes #51 

### LINKED ISSUE #51: Thumbs up and thumbs down indicators are missing from the playlist view.
Here's a side-by-side comparison of google play music in browser and in the desktop app.
capture

dst33nburgh:
The rating buttons weren't displaying in playlist view when the dark theme was turned on. This appears to fix it. It looks like the css property being set was overwriting the background image used to display the buttons.

This seemed like a really easy fix once I found what to change, so I'm skeptical as to whether or not this is the best way to fix it.


MarshallOfSound:
If it works, it works...

I've had issues exactly like this in the theme before where `background` overrode the background image and changing it to `background-color` worked.  I'm going to pull this down just to check it works then merge it :D

PR #53, opened 2015-11-26
Use HTTPS

DrTrust:


MarshallOfSound:
Seems fair enough, although for historical reasons I feel obligated to point out that google redirects all `http` requests to `https` anyway, this makes the result possibly quicker and more obvious in the code.

Issue #71, opened 2015-12-02
Bug: Custom Color not Applied


Devo7v:
When turning on the Custom Theme, the app is has a dark theme with green accents as the default.  If the user chooses another color (orange for instance) most of the accent colors change properly except for the check boxes in the settings and the word "Music"  in the "Google Play Music" slide out menu.  See the screenshot below.
Would it be possible to have the ability in the program to manually enter either a hex or rgb code for the custom color?
![screenshot 1](https://cloud.githubusercontent.com/assets/15945504/11545298/9b0037a2-9914-11e5-9aa4-74a77c30d1a6.jpg)

MarshallOfSound:
The RGB idea was suggested somewhere but can't seem to find it in an open issue so I'll leave this one open to track it.  

As for the "Music" in Google Play Music.  Theming that part it virtually impossible because it is by default orange and the only way I can change it is using CSS filters to adjust hue/saturation/brightness.  I'm looking into replicating the image as an SVG to make colours easier.

The checkboxes aren't actually part of the theme, I can make them part of it but green is a global signifier for good / active.  I think it is just confusing that the default highlight is green as well (The two greens are actually completely different)

TLDR: Hex color = Will do
           Music = Working on
           Checkboxes = Invalid

PR #76, opened 2015-12-03
Added text box for hex color input in the theme settings.

jogleasonjr:
For enhancement #71, inputting hex values for the theme color. If input validation fails, it defaults to black (#000000).

theme_hex

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/76)


Google Play Music.sln
```diff
@@ -17,8 +17,8 @@ Global
 		Release|x86 = Release|x86
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
-		{B2C52D42-4EEE-445C-AD6D-357D93397937}.Debug|Any CPU.ActiveCfg = Release|x86
```
MarshallOfSound: For future reference I don't like config changes like this coming in on a feature change.  It just doesn't look right.  We don't even use the "Any CPU" config because CEFSharp requires a specific platform target.

Leave it in for now as it doesn't do any harm, just an FYI :smile_cat:

Google Play Music/ColorWheel.cs
```diff
@@ -518,6 +519,51 @@ public static Color HslToRgb(HslColor hsl)
...
+        /// <returns>int value of color, or black if invalid</returns>
+
+        private static int HexToInt(string hex)
+        {
+
+            int rgb = 0;
+
+            if (!int.TryParse(hex.Replace("#", string.Empty), NumberStyles.HexNumber, null, out rgb))
+            {
+                //invalid hex
```
MarshallOfSound: What is the actual point of this if statement, appears to just be completely empty and perform no useful function

Google Play Music/ColorWheel.cs
```diff
@@ -518,6 +519,51 @@ public static Color HslToRgb(HslColor hsl)
...
+        /// <summary>
+        /// Convert hex value to int value. Filters leading # character.
+        /// </summary>
+        /// <param name="hex">Hex value of color</param>
+        /// <returns>int value of color, or black if invalid</returns>
+
+        private static int HexToInt(string hex)
+        {
+
+            int rgb = 0;
```
MarshallOfSound: Can the default RGB be configured to be the same as the default hexdecimal code in the default user settings file (found here)  

https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/master/Google%20Play%20Music/App.config#L28

Google Play Music/UpdateDialog.cs
```diff
@@ -80,20 +80,19 @@ public UpdateDialog(string changeLog, string oldVersion, string newVersion)
 
         private void InitializeComponent()
         {
-            // Auto generated by the designer
-            notNowButton = new MaterialRaisedButton();
-            this.updateButton = new MaterialRaisedButton();
-            this.changeLogText = new MaterialLabel();
+            this.notNowButton = new MaterialSkin.Controls.MaterialRaisedButton();
+            this.updateButton = new MaterialSkin.Controls.MaterialRaisedButton();
+            this.changeLogText = new MaterialSkin.Controls.MaterialLabel();
```
MarshallOfSound: It looks here as if your Visual Studio is configured differently to mine, all these namespace paths really aren't required as they are all in the "using".  Check your options so that this doesn't happen.

Google Play Music.sln
jogleasonjr: That's understandable. For what it's worth, I changed the configuration to Debug so that code optimization wouldn't prevent full debug sessions. I'll flip it back if you'd like.

Google Play Music/ColorWheel.cs
jogleasonjr: I forgot to implement color fallback when invalid hex was added. You requested this below. I implemented it in changes e846ced.

Google Play Music/ColorWheel.cs
jogleasonjr: Yes, done. See above comment. Cheers.

Google Play Music/UpdateDialog.cs
jogleasonjr: I can't find these settings. I attempted to mitigate it using Resharper, but even that looks different than what you had originally. Still learning Git here even though I'm well versed in .Net.

MarshallOfSound: In the future using a program like [Source Tree](https://www.sourcetreeapp.com/) might help analyzing what you have changed and choosing what to commit to Version Control.  I would pull your fork and fix it but then you won't get the glorious commit credit :+1: 

I reckon you should (if you are unfamiliar with git you may have to google these terms)
* Squash the two commits you have into one 
* Manually revert the changes made to this file
* Commit that manual reversion with the `--amend` flag to apply it to your latest commit.
* Force push it up to GitHub (`git push -f`)

Hopefully that makes sense :)

MarshallOfSound:
I'm sorry, I think a refactor I pushed yesterday caused some epic merge conflicts, can you look into those :crying_cat_face:

jogleasonjr:
@MarshallOfSound Yes, I'll get it sorted out this weekend. Assuming you want the hex entry merged in eventually.

Devo7v:
Any progress on this?

MarshallOfSound:
Closing due to progress on #171

PR #96, opened 2015-12-18
Removed Unnecessary Exceptions

Stevoni:
Added check so LastFM calls would no longer throw exceptions.
Added task killing pre-build to ensure the application will build if prematurely exited while debugging.

### DIFF
--- a/Google Play Music/Google Play Music.csproj
+++ b/Google Play Music/Google Play Music.csproj
@@ -320,6 +320,9 @@
   </Target>
   <Import Project="..\packages\cef.redist.x86.3.2357.1287\build\cef.redist.x86.targets" Condition="Exists('..\packages\cef.redist.x86.3.2357.1287\build\cef.redist.x86.targets')" />
   <Import Project="..\packages\CefSharp.Common.43.0.0\build\CefSharp.Common.targets" Condition="Exists('..\packages\CefSharp.Common.43.0.0\build\CefSharp.Common.targets')" />
+  <PropertyGroup>
+    <PreBuildEvent>taskkill /F /IM CefSharp.BrowserSubprocess.exe 2&gt;&amp;1 | exit /B 0</PreBuildEvent>
+  </PropertyGroup>
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it.
        Other similar extension points exist, see Microsoft.Common.targets.
   <Target Name="BeforeBuild">
--- a/Google Play Music/JSBound.cs
+++ b/Google Play Music/JSBound.cs
@@ -95,13 +95,10 @@ namespace Google_Play_Music
         // Fired from javascript when a different song has been playing for over half it's duration. (As according to last.fm's scrobbling spec)
         public void songScrobbleRequest(string song, string artist, string album, int timestamp)
         {
-            try
+            // Check last.fm user authenticated before attempting to "scrobble" the current song.
+            if (LastFM.user_key != null)
             {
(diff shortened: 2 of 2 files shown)


Google Play Music/Google Play Music.csproj
```diff
@@ -320,6 +320,9 @@
   </Target>
   <Import Project="..\packages\cef.redist.x86.3.2357.1287\build\cef.redist.x86.targets" Condition="Exists('..\packages\cef.redist.x86.3.2357.1287\build\cef.redist.x86.targets')" />
   <Import Project="..\packages\CefSharp.Common.43.0.0\build\CefSharp.Common.targets" Condition="Exists('..\packages\CefSharp.Common.43.0.0\build\CefSharp.Common.targets')" />
+  <PropertyGroup>
+    <PreBuildEvent>taskkill /F /IM CefSharp.BrowserSubprocess.exe 2&gt;&amp;1 | exit /B 0</PreBuildEvent>
```
MarshallOfSound: Can this somehow be moved to C# logic / calls in `Program.cs` so that if this ever happens to an installed version, users don't have to force close the process?

Google Play Music/JSBound.cs
```diff
@@ -95,14 +95,11 @@ public void goBig()
         // Fired from javascript when a different song has been playing for over half it's duration. (As according to last.fm's scrobbling spec)
         public void songScrobbleRequest(string song, string artist, string album, int timestamp)
         {
-            try
+            // Check last.fm user authenticated before attempting to "scrobble" the current song.
+            if (LastFM.user_key != null)
             {
                 new LastFM().scrobbleTrack(artist, song, timestamp).Wait();
             }
-            catch (Exception e)
```
MarshallOfSound: Although the code does indeed look nicer like this, the try catch is there to also catch the rare case where the user has changed their password while the application is still open and in this case the `user_key` will be rejected by last.fm and and exception will be thrown.

If you can come up with a way of handling that case then that would be great

MarshallOfSound:
Thanks @Stevoni  for the interest in this repo, the ideas behind the PR are sound, great fix for the damm `CefSharp.BrowserSubprocess.exe` problem.  I've had it many times and never find an automated fix.

Google Play Music/Google Play Music.csproj
Stevoni: This is a compilation issue caused by the process locking assemblies and migration to Program.cs will not correct the error. I'll be more clear in my next commit messages.

Additionally, glancing through the issues list didn't show similar problems with installed instances. If they are reported, the process killing can be added to Program.cs. I'm concerned for other applications that may be using the same process which will be terminated unintentionally, whereas the coding machine is in a more expected state than the installed machine.

Google Play Music/JSBound.cs
Stevoni: Ok, I didn't know the use case since I don't use LastFM. I guess I need to sign up for it and work on some of these use cases.

Stevoni:
I'm going to close this pull request and work on a solution to the LastFM errors to include better integration with the consumer with exceptions as a last resort.

Google Play Music/Google Play Music.csproj
MarshallOfSound: Mmmm, that makes sense, we don't want to be killing off other programs, people might not like that :smile:

MarshallOfSound:
@Stevoni fair enough

PR #100, opened 2015-12-20
Added Pre Build Event to Kill CEF Process

Stevoni:
Added a pre-build event to kill the CefSharp.BrowserSubprocess that will lock the assemblies when debugging is stopped unexpectedly.


MarshallOfSound:
Could you please squash this into a single commit and then I'll merge it, having two commits that cancel each other out is bad practice :+1:

Stevoni:
Although I disagree with reverting being bad practice, I'm not sure how to minimize this into a single commit? I'm still incredibly new to git (only proficient in SVN), do you have any suggestions on the git steps to complete?

MarshallOfSound:
To me the reverting thing is purely for historical reasons, looking down a commit log you shouldn't see a change was made and have to check further up that it was / wasn't reverted.

As for how to squash commits, the basic command is `git rebase master -i` but look into it more in this post

https://davidwalsh.name/squash-commits-git

Stevoni:
Well, I seemed to have fubared something in the process, yay for trial and error.

I'm going to close this and have a single commit pull request later today.

PR #102, opened 2015-12-21
Added Pre Build Event to Kill CEF Process 

Stevoni:
Added a pre-build event to kill the CefSharp.BrowserSubprocess that will lock the assemblies when debugging is stopped unexpectedly.


MarshallOfSound:
LGTM :+1: 

Sorry it took so long for this, I'm quite picky :smile_cat:

Stevoni:
Completely understand. This was an interesting and necessary learning experience in dealing with git.

Issue #93, opened 2015-12-14
Feature Request: Keyboard command to Find


hschneidman:
I often use Ctrl+F to find a song or artist from a long list. Would be nice if it was possible to do in the app. It doesn't need to be mapped to Ctrl+F necessarily.

FuzzAU:
I'm submitted a pull request #103 for this feature, I thought it was a good idea

MarshallOfSound:
@FuzzAU next time in your PR description say "Fixes #93" then when it gets merged this issue will automatically close :+1:

PR #103, opened 2015-12-21
Bind CTRL-F to jump to the find input box

FuzzAU:
This commit binds the Ctrl-F (commonly used find shortcut) to focus the find input box, allowing users to press Ctrl-F and begin searching. As per feature request on issue #93

### DIFF
--- a/Google Play Music/CoreMusicApp.Hooks.cs
+++ b/Google Play Music/CoreMusicApp.Hooks.cs
@@ -23,7 +23,6 @@ namespace Google_Play_Music
                         break;
                 }
             };
-            globalKeyMouseHook.KeyDown += new KeyEventHandler(gkh_KeyDown);
             globalKeyMouseHook.KeyUp += new KeyEventHandler(gkh_KeyUp);
             FormClosed += (send, e) =>
             {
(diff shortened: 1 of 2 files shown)


Google Play Music/CoreMusicApp.Hooks.cs
```diff
@@ -88,7 +88,12 @@ void gkh_KeyUp(object sender, KeyEventArgs e)
                         zoomInProgress = false;
                     }
                     break;
-
+                case Keys.F:
+                    if((ModifierKeys & Keys.Control) == Keys.Control)
```
MarshallOfSound: Please add the ` && ApplicationIsActivated()` from the above checks.  These keyboard hooks are global so in its current state Ctrl + F will go to the search box no matter what application you are currently in.

MarshallOfSound: Also although looking at it `ModifierKeys` will do the same job as my `controlDown` variable can we please be consistent, so either change yours to match mine, or change mine to match yours.  I don't mind either way as long as it works

Google Play Music/CoreMusicApp.cs
```diff
@@ -176,6 +176,11 @@ private void nextTrack()
             GPMBrowser.EvaluateScriptAsync("(function() {document.querySelectorAll('[data-id=forward]')[0].click()})()");
         }
 
+        private void findFocus()
+        {
+            GPMBrowser.EvaluateScriptAsync("(function() {document.getElementById('input').focus()})()");
```
MarshallOfSound: Just being picky now but could we update this selector to allow for Google's tendency to change ID's and classes of elements.  Something like `document.querySelectorAll('sj-search-box > input')[0].focus()` should be future proof enough I think.

MarshallOfSound:
Thanks for taking a shot at this, simple enough implementation I like it :cactus:

FuzzAU:
No problems, I've got another fix coming beyond the current one pushed up then it should be good to go :-)

FuzzAU:
I like your suggestions and implemented them as requested. :hamster:

MarshallOfSound:
LGTM 
Will merge when AppVeyor has done its thing :+1:

PR #105, opened 2015-12-21
Last FM Overhaul

Stevoni:
Added Log4Net for logging of LastFM log in attempts (only user name), errors, now playing, and scrobbles.
Reworked the LastFM integration to be a singleton to allow checking the user information before attempting to update or scrobble.
Replaced the System.Web Javascript JSON Deserializers with System.Runtime DataContractDeserializer. Can't remove the System.Web reference until migrating to .Net 4.5+ to use https://msdn.microsoft.com/en-us/library/system.net.webutility.urlencode(v=vs.110).aspx instead of HttpUtility.UrlEncode (LastFM.Integration/GenerateMethodParams).
Added album to the scrobble and song changed to correctly update widgets (in most cases).

Note: This overhaul continues to use the mobile style API rather than desktop style

FYI: Changing a password within the website will not cause errors with the updates as long as the previous authenticated user_key is used.

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/105)

### DIFF
--- a/Google Play Music/JSBound.cs
+++ b/Google Play Music/JSBound.cs
@@ -4,10 +4,11 @@ using System.Drawing;
 using System.IO;
 using System.Threading.Tasks;
 using System.Windows.Forms;
+using log4net.Core;
 
 namespace Google_Play_Music
 {
-    class JSBound
+    internal class JSBound
     {
         private CoreMusicApp mainForm;
 
@@ -16,18 +17,20 @@ namespace Google_Play_Music
             mainForm = parent;
             mainForm.FormClosing += (o, e) =>
             {
-                try {
+                try
+                {
                     if (alert != null)
                     {
                         alert.currentStep = 99999;
                     }
-                } catch (Exception)
+                }
+                catch (Exception)
                 {
                     // Just ignore it
                 }
             };
             // When the parent form is deactivated (focus is lost) we must activate the notification to ensure it stays in focus
-            
+
         }
 
         public void setInitialZoom()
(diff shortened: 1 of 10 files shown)


Google Play Music/JSBound.cs
```diff
@@ -116,66 +123,77 @@ public void songChangeEvent(string song, string album, string artist, string url
...
-                    Control.CheckForIllegalCrossThreadCalls = false;
                     alert.Dispose();
-                    Control.CheckForIllegalCrossThreadCalls = true;
                     alert = null;
-                };
+                }
+            }
+            else
+            {
+                // Based on previous intent, the Handle is required to be created but since the alert is closing, we need to dispose of the alert as soon as the handle is created for the alert.
```
MarshallOfSound: Good comments :+1:

Google Play Music/LastFM.Intergration.cs
```diff
@@ -11,97 +11,111 @@
...
-    class LastFM
+    internal class LastFM
     {
-        private string username;
-        private string password;
-        public static string user_key = null;
+        private string user_key = null;
+        private static LastFM instance;
 
         private const string API_KEY = "";
```
MarshallOfSound: Woo hoo, no API keys this time :+1:

MarshallOfSound:
Looking good, I'll take another look at it tomorrow morning (It's quite late here now) but looks like an easy merge :+1:

Stevoni:
Any update for this?

MarshallOfSound:
Sorry, kinda went on holiday. I'll look into this when I get back :+1:

MarshallOfSound:
OK, so I've given this a lot of thought and here is the feedback.  Basically it all comes down to the scope of the PR.

* First up there are now conflicts (oops) that need to be resolved
* Kill all the logging stuff, that should be a separate PR with logging added throughout the application, I'll probably handle this at some point
* Revert back to the old `Web Serialization` method of JSON handling unless you can explain why we need to change to this more convoluted method

Apart from those 3 things the base of the PR looks fine, just bleeding out of its scope a bit

PR #112, opened 2015-12-30
Change to using Nuget Automatic Package Restore

amaitland:
will resolve a lot of the problems restoring CefSharp

https://docs.nuget.org/consume/package-restore/migrating-to-automatic-package-restore
https://github.com/owen2/AutomaticPackageRestoreMigrationScript/blob/master/migrateToAutomaticPackageRestore.ps1


MarshallOfSound:
LGTM :+1:

PR #155, opened 2016-01-26
Store Playback Information in File

rynti:
### Description
This feature allows users to store information about the playback status
in a JSON-formatted file. The feature is disabled by default, users have
to manually opt-in through their desktop settings.

The playback information can be used to give external programs access to
the current song and playback status, for instance Twitch streamers
might want to display the currently played song.

These are the attributes of the playback information file:
- **isPlaying**: (bool) Whether or not a song is currently playing
- **song**: (string) The name of the current song
- **artist**: (string) The artist of the current song
- **album**: (string) The album of the current song
- **imageURL**: (string) An URL to the current song's album or song image

This is how the new setting looks in the desktop settings screen:
Image of the new Store Playback Information in File setting

### Changes

All changes should be self-explanatory, but just to make sure:
- Desktop Settings UI is changed to reflect the new option.
- PlayStatus class is introduced to manage the current playback status.
- New JavaScript interface setPlaybackStatus is introduced to
  update the "isPlaying" attribute of the playback information.
- The existing JavaScript interface songChangeEvent is used to update
  the song, artist, album, and imageURL attributes of the playback
  information.
- The FormClosing event of the main application is used to set the
  isPlaying attribute to "false" just before the player closes so the
  file reflects the state properly.

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/155)

### DIFF
--- a/Google Play Music/App.config
+++ b/Google Play Music/App.config
@@ -49,6 +49,12 @@
             <setting name="BetaStream" serializeAs="String">
                 <value>False</value>
             </setting>
+            <setting name="PublishPlaybackInformation" serializeAs="String">
+                <value>False</value>
+            </setting>
+            <setting name="PublishPlaybackInformationPath" serializeAs="String">
+                <value>%USERPROFILE%\gpm-status.json</value>
+            </setting>
         </Google_Play_Music.Properties.Settings>
     </userSettings>
   <runtime>
--- a/Google Play Music/CoreMusicApp.cs
+++ b/Google Play Music/CoreMusicApp.cs
@@ -57,6 +57,13 @@ namespace Google_Play_Music
             // Setup the Web Browser
             InitializeCEF();
 
+            FormClosing += (send, ev) =>
+            {
+                // Update PlayStatus when the app gets closed.
+                PlayStatus.Instance.information.isPlaying = false;
+                PlayStatus.Instance.Update();
+            };
+
             // Don't forget to save all our settings
             FormClosed += (send, ev) =>
             {
--- a/Google Play Music/Google Play Music.csproj
+++ b/Google Play Music/Google Play Music.csproj
@@ -229,6 +229,7 @@
     <Compile Include="CEF Modules\ResourceHandlerFactory.cs" />
     <Compile Include="CEF Modules\MenuHandler.cs" />
     <Compile Include="JSBound.cs" />
+    <Compile Include="PlayStatus.cs" />
     <Compile Include="Program.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Properties\Resources.Designer.cs">
--- a/Google Play Music/JSBound.cs
+++ b/Google Play Music/JSBound.cs
@@ -93,6 +93,12 @@ namespace Google_Play_Music
             });
         }
 
+        public void setPlaybackStatus(bool isPlaying)
+        {
+            PlayStatus.Instance.information.isPlaying = isPlaying;
+            PlayStatus.Instance.Update();
+        }
+
         public void setThumbbarToPlay()
         {
             mainForm.Invoke((MethodInvoker)delegate
@@ -135,6 +141,12 @@ namespace Google_Play_Music
         // Fired from javascript when a different song starts playing
         public void songChangeEvent(string song, string album, string artist, string url)
         {
+            PlayStatus.Instance.information.song = song;
+            PlayStatus.Instance.information.artist = artist;
+            PlayStatus.Instance.information.album = album;
+            PlayStatus.Instance.information.imageURL = url;
+            PlayStatus.Instance.Update();
+
             mainForm.Invoke((MethodInvoker)delegate
             {
                 // Update the title to show the current song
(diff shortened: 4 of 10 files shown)


MarshallOfSound:
Reviewed 10 of 10 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[Google Play Music/App.config, line 56 \[r1\]](https://reviewable.io:443/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/155#-K8zLROkX1jiQzjdJpM-:-K8zLROkX1jiQzjdJpM0:-827151724) ([raw file](https://github.com/marshallofsound/google-play-music-desktop-player-unofficial-/blob/2cdadded1377cd429fd7efd9c72d389726e12a12/Google Play Music/App.config#L56)):*
This seems a bit complicated for the average user, some people don't know about the %XXXX% file paths.  Can we have it default to empty, and then in the code if it is empty default it to the ACTUAL path that represents `%USERPROFILE%`

---

*[Google Play Music/JSBound.cs, line 147 \[r1\]](https://reviewable.io:443/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/155#-K8zLq7GJhp108qddWKi:-K8zLq7HFcnt8bO8yyI7:-349538192) ([raw file](https://github.com/marshallofsound/google-play-music-desktop-player-unofficial-/blob/2cdadded1377cd429fd7efd9c72d389726e12a12/Google Play Music/JSBound.cs#L147)):*
Instead of setting a much of attributes arbitrarily can we turn this into a method on `PlayStatus`.  Something like `PlayStatus.updateCurrentSong(song, artist, album, url)`

---

*[Google Play Music/PlayStatus.cs, line 49 \[r1\]](https://reviewable.io:443/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/155#-K8zM_syTDbUAr95zv09:-K8zM_sz1vzJNcsP2r40:-249300638) ([raw file](https://github.com/marshallofsound/google-play-music-desktop-player-unofficial-/blob/2cdadded1377cd429fd7efd9c72d389726e12a12/Google Play Music/PlayStatus.cs#L49)):*
This line has an uncaught exception when the file path is completely invalid.  For instance  
`ZZ:::\\\\//asfloib/asf\a`
We should catch bad file paths in the settings dialog and warn the user there2

---

*[Google Play Music/SettingsDialog.Designer.cs, line 97 \[r1\]](https://reviewable.io:443/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/155#-K8zMy_tArmrQ2s_xaz2:-K8zMy_tArmrQ2s_xaz3:-692033021) ([raw file](https://github.com/marshallofsound/google-play-music-desktop-player-unofficial-/blob/2cdadded1377cd429fd7efd9c72d389726e12a12/Google Play Music/SettingsDialog.Designer.cs#L97)):*
Funky font changes, they should be Roboto....

I wonder when they changed

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/155)*

MarshallOfSound:
I'm still considering whether or not this should be merged into this program.  It is definitely an *advanced* feature as most (98%) of users wouldn't use it.

Also please keep in mind that that I have a contributing.md file https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/master/CONTRIBUTING.md#pull-requests

You can already get the current playing song from the title of the application and I know a few people have figured out how to use that in their twitch streams.

Fix up the things I commented on and I'll think about this PR's purpose in the overall app

rynti:
I completely understand your perspective and totally agree that this is
something for advanced users only. The issue is that without this
feature, there is currently no way of figuring out whether the player is
currently actually playing a song or not.

However, let me propose this:
How about I remove the path setting completely and only introduce the
"Store Playback Information in File" checkbox. That should suffice
completely and makes the feature less intrusive in the Desktop settings
(One small checkbox doesn't harm anyone I think).

On a later occasion we could also implement a question mark icon next to
the checkbox, explaining what the feature would do, although for now it
would probably be enough if we explain the feature somewhere in the
documentation on the GitHub page.

Let me know your thoughts, the removal of the path setting would also
get rid of the issues you had of average users not having an idea what
the %XXXX% file paths mean.

*(On a sidenote: I did read the contributing.md file beforehand, was
there any part of it in particular you were missing in my pull request?)*

MarshallOfSound:
In regards to contributing.md

> **Please ask first** before embarking on any significant pull request (e.g. implementing features)

Just to prevent a discussion like this happening in a PR, next time an issue for discussion before PR would be great :smile_cat: 

Removing the path would be a step in the right direction but I don't want the settings box (it's small as it is) to get bigger and more flooded.

Just a thought, why does this even need to be an option.  Why not just spit out the file no matter what to the APPDATA directory.  It won't even be a kilobyte so no harm for those that don't want it.

Just remove the setting and just make it do it ?

rynti:
About the part in the contributing.md: I thought the "ask first" was primarily meant for huge features so people wouldn't waste time implementing something that would not even make it into the final application. Since I needed this functionality anyway, I just thought I'd also create a pull request so others could benefit as well :)

The complete removal of the setting seems like a good solution, I'll get to it and create a new pull request when it's done, thanks for the feedback. I've created an issue for further discussion: #156

camjac251:
Did you ever implement this with OBS? Like having OBS parse the JSON, showing the artist - song, and album art?

PR #209, opened 2016-03-05
linux support

devTristan:
[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/209)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -105,7 +105,7 @@ gulp.task('transpile', ['clean-internal'], () => {
       presets: ['es2015'],
     }))
     .on('error', (err) => { console.error(err); }) // eslint-disable-line
-    .pipe(replace(/process\.env\.(.+)\;/gi, (envCall, envKey) => {
+    .pipe(replace(/process\.env\.([^\.]+)\;/gi, (envCall, envKey) => {
       return `'${process.env[envKey]}'`;
     }))
     .pipe(gulp.dest('./build/'));
--- a/src/inject/generic.js
+++ b/src/inject/generic.js
@@ -5,7 +5,7 @@ import SettingsController from '../main/utils/Settings';
 global.Settings = new SettingsController();
 Settings.uncouple();
 
-require(`./${process.platform}`);
+require(`./${process.platform === 'win32' ? 'win32' : 'darwin'}`);
 
 const waitForBody = setInterval(() => {
   if (document.body) {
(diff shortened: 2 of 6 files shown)


Atraii:
I'm getting an error on this build: 
`codesign <snip> 
/bin/sh: command not found`

What dependency should be in place to get the codesign package on linux? (Also it appears that circleci needs a newer version of node.)

flakeparadigm:
When building and running it from the source (no installer being created), everything I've tested seems to be working except the media control keys. Not sure if that's within the scope of this PR, but figure'd I'd leave a note about it.

Setup: Ubuntu 15.10 with Node 4.3

src/inject/generic.js
```diff
@@ -5,7 +5,7 @@ import SettingsController from '../main/utils/Settings';
 global.Settings = new SettingsController();
 Settings.uncouple();
 
-require(`./${process.platform}`);
+require(`./${process.platform === 'win32' ? 'win32' : 'darwin'}`);
```
MarshallOfSound: I'm not sure this is what we want?

This import darwin (osx) based injections when building for `linux`.  Although this may work now it is counter intuitive, I think we should leave as is and have a `"linux"` folder that might possibly import things from the `darwin` folder to avoid code duplication.  Then it makes more sense

src/main/utils/appdir.js
```diff
@@ -0,0 +1,18 @@
+let appdir;
+
+if (process.env.APPDATA) {
```
MarshallOfSound: You need to have `const process = process.env` here like before or my environment variable transform breaks this code

gulpfile.js
```diff
@@ -105,7 +105,7 @@ gulp.task('transpile', ['clean-internal'], () => {
       presets: ['es2015'],
     }))
     .on('error', (err) => { console.error(err); }) // eslint-disable-line
-    .pipe(replace(/process\.env\.(.+)\;/gi, (envCall, envKey) => {
+    .pipe(replace(/process\.env\.([^\.]+)\;/gi, (envCall, envKey) => {
```
MarshallOfSound: My regexing skillz aren't the best but this doesn't change anything right?  It simply blocks

`process.env..SOMETHING`?

MarshallOfSound:
Closing this as we got linux support in another PR

Issue #212, opened 2016-03-05
OSX issues


ValentinDenis:
Just listing here my remarks made on the reddit post:
 - better tray icon (and a white one for dark mode)
![image](https://cloud.githubusercontent.com/assets/6633654/13549710/14986382-e30c-11e5-90fc-1bc3a7d5d5c0.png)
 - Full screen
 - Voice commands not working for me

snigel1:
Also, setting hotkeys does not seem to work

wiiareonfire:
when "built in audio" is selected as the playback device, changing the system volume does not change playback volume

close, minimize, and full screen are on the right side of the window bar (typically on left side in mac applications)

MarshallOfSound:
@ValentinDenis Voice commands are off by default, you have to enable them from the "General" section of the settings menu.

I'll take a look at that PR real quick

MarshallOfSound:
Closing as the icon has been fixed, the voice controls is user error and full screen is incoming in another PR

PR #221, opened 2016-03-05
Fix Mac Fullscreen

clatour:
Looking at the [electron docs](http://electron.atom.io/docs/latest/api/browser-window/), `{fullscreenable: true}` (which should be true by default) should work in the place of this in the index.js. I attempted to modify the `baseConfig` with this and it didn't seem to work. I haven't looked into electron itself, but using `mainWindow.setFullScreenable(true); ` as a workaround after init seems to work for me on OS X 10.11.3.

Adds a separate maximize toggle path for `darwin`. 

`WindowManager.getByInternalID(windowID)` calls removed since the `window`/`windowID` doesn't change. Not super related to the feature pull.

Addresses part of #212.

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/221)

### DIFF
--- a/src/main/features/core/controlBar.js
+++ b/src/main/features/core/controlBar.js
@@ -4,10 +4,19 @@ Emitter.on('window:minimize', (event, windowID) => {
 
 Emitter.on('window:maximize', (event, windowID) => {
   const window = WindowManager.getByInternalID(windowID);
-  if (window.isMaximized()) {
-    WindowManager.getByInternalID(windowID).unmaximize();
+
+  if (process.platform === 'darwin') {
+    if (window.isFullScreen()) {
+      window.setFullScreen(false);
+    } else {
+      window.setFullScreen(true);
+    }
   } else {
-    WindowManager.getByInternalID(windowID).maximize();
+    if (window.isMaximized()) {
+      window.unmaximize();
+    } else {
+      window.maximize();
+    }
   }
 });
 


MarshallOfSound:
I like this, let me just see what happened on travis and I'll probably merge

src/index.js
```diff
@@ -58,6 +58,9 @@ import handleStartupEvent from './squirrel';
     mainWindow = new BrowserWindow(generateBrowserConfig());
     global.mainWindowID = WindowManager.add(mainWindow, 'main');
 
+    // Required for 'darwin'
+    mainWindow.setFullScreenable(true);
```
MarshallOfSound: This shouldn't be required, can we try without it?

clatour: Think I found the issue: https://github.com/atom/electron/blob/v0.36.10/atom/browser/native_window_mac.mm#L490-L491

Looks like we need to set `{fullscreen: true}` as well (or in place of `fullscreenable`) in `src/main/configureBrowser.js` if we want it to Just Work without this line. I'm of the opinion that the app shouldn't default itself to fullscreen, but your call if you want to go that route. In that case this line can be removed.

Can try to patch electron such that `fullscreenable`can be set (via config options) despite `fullscreen` being `false` (and probably should override `maximizable` anyways). I'm the type of user that hates the maximize behavior on Mac apps in favor of "real" fullscreen, but there might be justification for having it the way it is. Otherwise this is probably **_the_** workaround, and maybe we can help them out with the docs.

Trying to poke around in https://github.com/atom/atom for reference into how they handle it, but haven't found it so far.

MarshallOfSound: I think I changed fullscreen to be true recently

clatour: Ah sorry, didn't have an up-to-date copy. Electron docs say by default it's false, but removing it seems to negate the need for this line. I was under the impression that setting it to true would open the app in fullscreen.

MarshallOfSound: Hm, strange.  Electron docs are not the most precise things in the world.

If it works it works :+1:

clatour:
Workaround removed & rebased.

src/main/features/core/controlBar.js
```diff
@@ -4,10 +4,19 @@ Emitter.on('window:minimize', (event, windowID) => {
 
 Emitter.on('window:maximize', (event, windowID) => {
   const window = WindowManager.getByInternalID(windowID);
-  if (window.isMaximized()) {
-    WindowManager.getByInternalID(windowID).unmaximize();
+
+  if (process.platform === 'darwin') {
+    if (window.isFullScreen()) {
```
jawshooah: Could trim this down to one line with `window.setFullScreen(!window.isFullScreen())`

MarshallOfSound:
Closing this in favour of #386 (completely native windows borders instead of replicating behaviour)

PR #226, opened 2016-03-05
Initial Linux support for GPMDP

alberthdev:
I attempted to implement Linux support for GPMDP!

What I did:
 * Modify `Settings.js` and `PlaybackAPI.js` to use `os.homedir()` if the platform is not Darwin. `/var/local` is usually locked by the system, and GPMDM shouldn't need admin access to run! On Linux, this now stores configuration in `/home/USER/GPMDP_STORE` (or whatever home directory is defined).
  * You may want to consider switching `environment.HOME` to `os.homedir()` to keep things consistent!
 * Create empty `src/main/features/linux/` module to avoid import error.
 * Copy `inject`'s `darwin` platform to module to `src/inject/linux/` to avoid import error. (I think this was the best fit, at least until you get notifications implemented.
 * Add Linux binary and package building.
 * Did a rough implementation of your package upload script - renamed original to have `-mac` suffix, and added a new `-linux` script that uploads both the x86 and x64 builds.
 * Added Linux building to Travis-CI. To prevent strange upload attempts, I commented out the line that runs the Linux upload script. Simply change it to enable automatic builds!

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/226)


src/inject/linux/index.js
```diff
@@ -0,0 +1 @@
+import './silent_Notification';
```
MarshallOfSound: Could change this to be `import '../darwin/slient_Notification';  Then we don't need to copy paste

alberthdev: Better yet, could we move silent_Notification to the root directory? Since it's not quite platform-specific anymore:

`import '../silent_Notification';`

PR update incoming...

MarshallOfSound:
Overall this looks like a really good starting place!!!

Thanks so much for this.  I'm just pulling down Ubuntu ISO's so I can check a bunch of things work now

alberthdev:
Thanks! Regarding checks:

 * CircleCI seems to like to build Mac stuff... on non-Mac platforms. Do you want me to fix that for you?
 * Travis-CI is failing since `KEY_PASSWORD` isn't defined (or isn't correct?) for PR builds. This is normal though since if we were able to see that, I could easily expose your password... so PR incoming to fix that!

alberthdev:
Actually, just ran CircleCI myself... you may need to change your settings to avoid building anything. (You might need to remove some commands from CircleCI's settings...)

MarshallOfSound:
Yep, can you fix the KEY_PASSWORD thing for darwin builds aswell :+1: Then we can get green builds going.

I turned of circleci a few minutes ago so it shouldn't bother us any more :)

MarshallOfSound:
I just setup the linux artifact server

.travis.yml
```diff
@@ -18,7 +19,10 @@ branches:
     - website
 
 script:
-  - npm run test && ./sig/import.sh && npm run make:darwin && node ./sig/publish.js
+  - npm run test
+  - npm run build
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]] && [[ $TRAVIS_PULL_REQUEST == "false" ]]; then ./sig/import.sh && npm run make:darwin && node ./sig/publish-mac.js; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then npm run make:linux && echo Success; fi #&& node ./sig/publish-linux.js; fi
```
MarshallOfSound: Surely this line needs this aswell?  
` && [[ $TRAVIS_PULL_REQUEST == "false" ]];`

MarshallOfSound:
@alberthdev It also looks like we will have to disable the "minimize to tray" functionality on linux.  Tray icon support is varied and random

.travis.yml
alberthdev: Which line are you referring to? The normal build line doesn't invoke any of the `make:XXXX` rules, and the Linux line doesn't need it since it doesn't use any signing whatsoever.

alberthdev:
Hmmm... minimize to tray does work for me. When you minimize it doesn't actually minimize to tray, but closing it will.

.travis.yml
MarshallOfSound: Ah never mind, it looks like the publish command is commented out *bows out*

MarshallOfSound:
On my clean ubuntu build the tray icon doesn't appear at all?

alberthdev:
Just looked around - according to [this](https://github.com/atom/electron/blob/master/docs/api/tray.md) you may need to install libappindicator1 to get the icon to show up.

MarshallOfSound:
Can we somehow check if that is installed on linux and change behaviour accordingly?

Check if a file exists inside `/etc`?

alberthdev:
Hmm.... you could definitely check for the existence of `/usr/lib/libappindicator.so.1`. Better yet, check for `/usr/lib/libappindicator*` existence too. This seems to only be the case for Unity/GNOME 3.x desktops though - older desktops (or sensible ones!) have implemented support for both the regular tray icon and the new Appindicator tray.

Running `lsb_release -i` could give you information about the distro, which you can use to make things work.

All in all, I think the best way to detect is if the users detect it themselves. On first startup, show them the tray icon, and ask the user if they see it in their taskbar. If they do, they can click "Enable Minimize to Tray". If not, they can click "Disable Minimize to Tray". No need for messy detection! (Especially since it does work on "legacy tray" systems...)

MarshallOfSound:
Sounds good, we can just add another modal thingy to the main `index.html` and pop it up if / when we want to :)

To be honest I'll merge this whenever and continue some touchups here and there in master.  Just let me know when you think you are done with it.

alberthdev:
Awesome! I think I'm pretty much done with it - if you want me to go ahead and enable Linux binary upload support, I can do that as well before the merge!

MarshallOfSound:
Yep, enable the upload and then I'll merge.  Then we should have an initial linux build up and running

alberthdev:
Ohh, that's why you were asking for that condition. Latest commit should now have that! (Hopefully nothing is broken...)

MarshallOfSound:
I'll just wait for travis to finish (better safe than sorry) then I'll merge

Thanks so much for this boost towards linux :+1:

MarshallOfSound:
Travis looking good merging now

alberthdev:
@MarshallOfSound 
No problem! I'm a big Linux fan, and since this was already on a cross-platform toolkit, most of the hard work is done - just needed a kick or two to get it to Linux. Thanks for all of your hard work!

screenshot - 03052016 - 08 28 22 pm
screenshot - 03052016 - 08 29 23 pm

PR #244, opened 2016-03-06
rpm and deb packages round two

dylanlesko:
Added options to make packages for more platforms.
npm run make:deb
npm run make:rpm

npm run make:linux will build both assuming you have apt and rpm.

Can probably shorten the code, a bit of redundant stuff going on, but wouldn't change performance.

Packages should work for Ubuntu/Mint/Fedora/Debian. Hoping to get one for Arch. Will probably add some testing for the linux platforms. Or might look through the issues.

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/244)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,9 @@ OS Support
 * Windows 7 or later
 * Mac OSX 10.8.0 or later
 * Ubuntu 14.04 or later
+* Mint
+* Fedora
+* Debian
 
 **Note: We do not current have automated builds for Ubuntu (PR's are welcome)**
 
@@ -76,6 +79,12 @@ npm run make:win
 
 # Mac OSX
 npm run make:darwin
+
+# Ubuntu
+npm run make:deb
+
+# Fedora
+npm run make:rpm
 ```
 
 All releases will be signing with my Code Signing Certificates (Authenticode on Windows and Codesign on OSX)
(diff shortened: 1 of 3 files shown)


MarshallOfSound:
Can we put the deb/rpm built installers into different folders (not just `./dist/installers`)

How about
* `./dist/installers/fedora`
* `./dist/installers/debian`

Then once merged I will modify other make processes to output to `darwin` and `win32` directories

dylanlesko:
Alright, made that requested change and have committed it to my fork. 

* `./dist/installers/installers.zip` is still made, and has every package in it.

MarshallOfSound:
That zip is only made on `make:linux` though right?  Or is it made on any linux related `make`?

dylanlesko:
I will double  check, but I think it will do it on `make:deb` and `make:rpm` as well. I could change that if you wanted, but I figured you wanted it that way since you had it zip before there were multiple linux builds. There are still the 32 and 64 bit versions that get zipped.

dylanlesko:
Just finished running `make:deb`.

It creates a zip `./dist/installers/debian/installers.zip` like I expected. `./dist/installers/installers.zip` isn't created or altered.

MarshallOfSound:
Looking good, thanks for the contrib!

PR #272, opened 2016-03-06
Increase settings window height to fit contents.

### LINKED ISSUE #240: Settings Window Cuts Off Bottom Two Options in Settings
It isn't clear to the user this is a scrollable area, at least on Mac OS.
screen shot 2016-03-06 at 5 55 07 pm

Fix would be to force a scroll bar, or ideally have the default window size show all the available settings.

tonygambone:
There are two tabs in the settings that must scroll (General and Hotkeys, though Hotkeys doesn't have any hidden content).

This fixes #240

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/272)

### DIFF
--- a/src/main/features/core/desktopSettings.js
+++ b/src/main/features/core/desktopSettings.js
@@ -8,7 +8,7 @@ export const showDesktopSettings = () => {
   }
   const desktopSettings = new BrowserWindow({
     width: 800,
-    height: 400,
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
Simple enough

LGTM :+1:

Issue #276, opened 2016-03-06
Open on left-click of the tray icon


MightyPork:
Got it working finally and it's actually pretty good!

But the tray icon is annoying. All Linux apps open on left-clicking the tray icon. (Well, apart for Steam, which is weird in many ways).

Is it possible to add left-click action to open the player? It'd be a ton easier to use.

Also tray buttons for prev / next / pause would be SUPER USEFUL.

MightyPork:
Hmm I found how to do this. Do you accept pull requests? Left click on the icon should toggle the main window, ideally.

MightyPork:
closing, it's all now in the PR

PR #278, opened 2016-03-06
Playback & like buttons in tray menu. Toggle minimize on tray left click.

MightyPork:
This pull request addresses my issue #276.

It adds some useful buttons to the tray menu, and also left-clicking the tray button toggles the app window, like in almost all Linux apps (before it was double-click and could not close, only open).

I hope it'll be ok on Windows and Mac, can't try it myself.

screenshot_20160306_195211

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/278)

### DIFF
--- a/src/main/features/core/tray.js
+++ b/src/main/features/core/tray.js
@@ -27,6 +27,29 @@ const setContextMenu = () => {
         mainWindow.show();
       },
     },
+    { type: 'separator' },
+    {
+      label: 'Play / Pause',
+      click: () => Emitter.sendToGooglePlayMusic('playback:playPause'),
+    },
+    {
+      label: 'Previous Track',
+      click: () => Emitter.sendToGooglePlayMusic('playback:previousTrack'),
+    },
+    {
+      label: 'Next Track',
+      click: () => Emitter.sendToGooglePlayMusic('playback:nextTrack'),
+    },
+    { type: 'separator' },
+    {
+      label: 'Thumbs Up',
+      click: () => Emitter.sendToGooglePlayMusic('playback:thumbsUp'),
+    },
+    {
+      label: 'Thumbs Down',
+      click: () => Emitter.sendToGooglePlayMusic('playback:thumbsDown'),
+    },
+    { type: 'separator' },
     {
       label: 'Audio Device',
       submenu: audioDeviceMenu,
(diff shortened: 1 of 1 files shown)


MightyPork:
The 'darwin' part needs testing & may be wrong, it's my best guess. Tested only on Linux in Xfce4.

src/main/features/core/tray.js
```diff
@@ -40,9 +63,35 @@ const setContextMenu = () => {
...
 appIcon.setToolTip('Google Play Music');
-appIcon.on('double-click', () => {
-  mainWindow.setSkipTaskbar(false);
-  mainWindow.show();
+
+
+// Iconified     -> show, to foreground
+// In background -> to foreground
+// In foreground -> iconify
+appIcon.on('click', () => {
```
MarshallOfSound: Can we extract this to it's own function `toggleAppVisiblitiy` or something similar?

Then have the same action on single click and double click?

src/main/features/core/tray.js
```diff
@@ -40,9 +63,35 @@ const setContextMenu = () => {
...
+  if (process.platform === 'darwin') {
+    // Special code for OS-X - Not tested!
+    if (!mainWindow.isVisible()) {
+      mainWindow.setSkipTaskbar(false);
+      mainWindow.show();
+    } else {
+      if (PlaybackAPI.isPlaying()) {
+        Emitter.sendToGooglePlayMusic('playback:playPause');
```
MarshallOfSound: This seems wrong but I can't test until I get home.  We shouldn't be pausing the player by clicking the tray icon

src/main/features/core/tray.js
```diff
@@ -40,9 +63,35 @@ const setContextMenu = () => {
...
+      }
+      mainWindow.hide();
```
MarshallOfSound: We shouldn't be hiding the player when clicking the tray icon without checking `Settings.get('minToTray`, true)`

src/main/features/core/tray.js
```diff
@@ -40,9 +63,35 @@ const setContextMenu = () => {
...
+    }
+  } else {
+    // Windows, Linux
+    if (mainWindow.isMinimized() || !mainWindow.isFocused()) {
+      // Window in tray, or not focused
+      mainWindow.setSkipTaskbar(false);
+      mainWindow.show();
+      mainWindow.focus();
+    } else {
+      mainWindow.minimize();
```
MarshallOfSound: We shouldn't be hiding the player when clicking the tray icon without checking `Settings.get('minToTray`, true)`

src/main/features/core/tray.js
MightyPork: I copied this from some other place where it's done like this, but probably it doesn't make sense indeed. Will remove it.

MarshallOfSound: Ah, I see where it came from.  Yeah, this isn't the behavior we want here :+1:

MightyPork:
Okay I think all the problems are more or less fixed, but there is something funny going on with the window.

Not sure if it's a Xfce4 quirk, or happens on other platforms:
- Close window with the cross button to tray
- Open from tray

It should get focused, but isn't. I haven't figured out why.
Currently I print an error message, and pretend it was focused properly.

MightyPork:
Okay actually I'll split this into two pull requests, one for the menu, one for the icon clicking.

So you can merge the menu, and we can figure this out later.

Issue #274, opened 2016-03-06
Global Hotkeys for Volume Control


jwmann:
Something that I've seen missing from *EVERY* Google Music Player I've used is global volume control.
Yes, you can change the volume using `-` and `=` while inside the player but that requires me to find the app and focus it to control it.

It would insanely appreciated if a Global Volume control hotkeys could be added.

Thank you.

MightyPork:
When my pull request gets merged, this will be possible ;-)

jwmann:
Beautiful :+1: 
Thank you very much! :smile:

PR #284, opened 2016-03-06
Added hotkeys for volume control

MightyPork:
Solves issue #274

Hotkeys can now be assigned for volume control.

Also solves a problem where new hotkeys could not be added to the `customHotkeys` dict, if user already saved the settings.

### DIFF
--- a/src/main/features/core/keyboardShotcuts.js
+++ b/src/main/features/core/keyboardShotcuts.js
@@ -17,14 +17,21 @@ globalShortcut.register('MediaStop', () => {
   Emitter.sendToGooglePlayMusic('playback:stop');
 });
 
-const customHotkeys = Settings.get('hotkeys', {
+const customHotkeysTemplate = {
   playPause: null,
   stop: null,
   previousTrack: null,
   nextTrack: null,
   thumbsUp: null,
   thumbsDown: null,
-});
+  increaseVolume: null,
+  decreaseVolume: null,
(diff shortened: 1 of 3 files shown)


src/public_html/desktop_settings.html
```diff
@@ -126,6 +126,16 @@ <h5 class="theme-text">Choose your theme highlight color</h5>
                 <label for="last_name">Thumbs Down Current Song</label>
               </div>
             </div>
+            <div class="row">
```
MarshallOfSound: I didn't think that this would work without adding extra inject JS to the settings tab, turns out I wrote some funky awesomeness that handles it :+1:  to me :smile:

src/main/features/core/keyboardShotcuts.js
```diff
@@ -17,14 +17,21 @@ globalShortcut.register('MediaStop', () => {
...
   thumbsUp: null,
   thumbsDown: null,
-});
+  increaseVolume: null,
+  decreaseVolume: null,
+};
+
+const userHotkeys = Settings.get('hotkeys', {});
+
+const customHotkeys = _.extend(customHotkeysTemplate, userHotkeys);
```
MarshallOfSound: Solid catch :+1:

MarshallOfSound:
Looking good, solid PR

Thanks!

:cactus:

PR #287, opened 2016-03-06
use XDG_CONFIG_HOME if present

Polymetric:
fixes the issue on Linux where the STORE directory in placed in a users home dir.

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/287)

### DIFF
--- a/src/main/utils/PlaybackAPI.js
+++ b/src/main/utils/PlaybackAPI.js
@@ -7,7 +7,7 @@ const environment = process.env;
 
 class PlaybackAPI {
   constructor() {
-    const DIR = (environment.APPDATA ||
+    const DIR = (environment.APPDATA || environment.XDG_CONFIG_HOME ||
       (process.platform === 'darwin' ? environment.HOME + '/Library/Preferences' : os.homedir())) +
       '/GPMDP_STORE';
     this.PATH = `${DIR}/playback.json`;
--- a/src/main/utils/Settings.js
+++ b/src/main/utils/Settings.js
@@ -8,7 +8,7 @@ const environment = process.env;
 
 class Settings {
   constructor(jsonPrefix, wipeOldData) {
-    const DIR = (environment.APPDATA ||
+    const DIR = (environment.APPDATA || environment.XDG_CONFIG_HOME ||
       (process.platform === 'darwin' ? environment.HOME + '/Library/Preferences' : os.homedir())) +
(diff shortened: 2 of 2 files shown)


src/main/utils/PlaybackAPI.js
```diff
@@ -7,7 +7,7 @@ const environment = process.env;
 
 class PlaybackAPI {
   constructor() {
-    const DIR = (environment.APPDATA ||
+    const DIR = (environment.APPDATA || environment.XDG_CONFIG_HOME ||
```
MarshallOfSound: This should only take effect on linux correct?

Can we move it to be this  
```js
const DIR = (environment.APPDATA ||
    (process.platform === 'darwin' ? environment.HOME + '/Library/Preferences' : (environment.XDG_CONFIG_HOME || os.homedir())) +
    '/GPMDP_STORE';
```

Polymetric:
That's another way to do it, although it feels a little messier. Your pick. The XDG_CONFIG_HOME env var will only be preset on systems that respect the freedesktop directory guidelines, i.e. most flavors of Linux.

MightyPork:
I just checked and I actually don't have XDG_CONFIG_HOME at all.

So probably good idea to make `~/.GPMDP_STORE` or `~/.config/GPMDP_STORE` a fallback if the variable is missing on Linux.

MightyPork:
@Polymetric any plays to complete this? If not close it n I'll redo it with the improvements, want it merged

jacobmischka:
Is there a reason we can't just use elecron's native method for getting a folder like this? https://github.com/atom/electron/blob/master/docs/api/app.md#appgetpathname `userData`

MarshallOfSound:
@jacobmischka Goddammit, I should have known there would be a method for that....

But now for backwards compatablity sake we shouldn't change it :cry:

jacobmischka:
I really think we should

MarshallOfSound:
So backwards compat check and copy?  I could go with that.  Will @Polymetric change this PR or will @MightyPork take over or does someone else what to crush this out?

jacobmischka:
If no one else has claimed it in an hour or so I'll do it.

Polymetric:
I really ought to turn on more notifications from GitHub. Thanks for taking this @jacobmischka

PR #290, opened 2016-03-06
Use hidden directory for config on Linux

MightyPork:
This PR moves the config directory to `~/.GPMDP_STORE` - a hidden directory, as is customary on Linux.

Linux users go to great lengths to make the home dir tidy, and then there's this, in all caps and visible!
Small downside is that Linux users will lose their config with this update, but it's worth it.

I have also abstracted the config directory resolution to it's own file. Please check if it's still working OK on Win and Mac.

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/290)

### DIFF
--- a/src/main/utils/PlaybackAPI.js
+++ b/src/main/utils/PlaybackAPI.js
@@ -1,14 +1,9 @@
 import fs from 'fs';
-import mkdirp from 'mkdirp';
-import config_dir from './config_dir';
-
-const os = require('os');
-
-const environment = process.env;
+import configDir from './configDir';
 
 class PlaybackAPI {
   constructor() {
-    this.PATH = `${config_dir}/playback.json`;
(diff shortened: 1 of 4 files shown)


MightyPork:
#287 probably solves this better, but I think moving this logic to it's own file is a good idea.

Polymetric:
Why not both? ;) I do think that respecting the XDG directories (putting it in `~/.config`) is a good plan.

MightyPork:
well that's what I meant :P
feel free to steal that piece from my PR, as it'll get closed anyway

MarshallOfSound:
I agree with @Polymetric 
I like the idea of moving the "config/data" file creation to a separate file.  It will make the logging work I'm about to start much easier.
If @Polymetric can steal that file and implement it into his PR I'll merge that one :+1:

PR #291, opened 2016-03-06
Option to disable the JSON API

MightyPork:
I'm afraid the way the API is done will cause harm to SSD drives and similar with limited number of write cycles. It writes the file every second!

Also not many people will use this, so why not have an option to disable it?

I have added a config key `"enableJSONApi"`, which is true by default, but can be disabled by editing the config file.

There is no toggle in the settings page, but I don't think it's necessary. It can be added later.

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/291)

### DIFF
--- a/src/main/utils/PlaybackAPI.js
+++ b/src/main/utils/PlaybackAPI.js
@@ -57,7 +57,9 @@ class PlaybackAPI {
   }
 
   _save() {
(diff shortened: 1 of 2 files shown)


MarshallOfSound:
Fair enough, pretty sure you're overestimating the effects of a single file being written.  But I'll take the PR as it doesn't really change much :+1:

PR #297, opened 2016-03-07
Main window toggle on tray icon click.

MightyPork:
This is the click-toggle part of #278.

I tried keeping both 'click' and 'double-click' handlers, but it didn't work, so only the 'click' handler is currently enabled. Maybe it could be configurable, but 'click' feels more comfortable (imo).

There is one glitch (which however **is not new**). Please see the comments.<br>
I've tried some hacks and nothing worked, so we can probably just let it be. It's not that bad.

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/297)

### DIFF
--- a/src/main/features/core/tray.js
+++ b/src/main/features/core/tray.js
@@ -75,7 +75,7 @@ function toggleMainWindow() {
 }
 
 appIcon.setToolTip('Google Play Music');
-//appIcon.on('double-click', toggleMainWindow);
+// appIcon.on('double-click', toggleMainWindow);
 appIcon.on('click', toggleMainWindow);
 
 
--- a/src/main/features/core/tray.js
+++ b/src/main/features/core/tray.js
@@ -95,7 +95,10 @@ function toggleMainWindow() {
 
 appIcon.setToolTip('Google Play Music');
 appIcon.on('click', toggleMainWindow);
-appIcon.on('double-click', toggleMainWindow);
+
+/* does not work if the 'click' handler is present */
+//appIcon.on('double-click', toggleMainWindow);
+
 
 // DEV: Keep the icon in the global scope or it gets garbage collected........
 global.appIcon = appIcon;
--- a/src/main/features/core/tray.js
+++ b/src/main/features/core/tray.js
@@ -61,7 +61,7 @@ function toggleMainWindow() {
 appIcon.setToolTip('Google Play Music');
 
 switch (process.platform) {
-  case 'darwin': // actually means OS-X
+  case 'darwin': // <- actually means OS-X
     // No toggle action, use the context menu.
     break;
   case 'linux':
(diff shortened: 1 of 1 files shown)


src/main/features/core/tray.js
```diff
@@ -39,11 +39,66 @@ const setContextMenu = () => {
...
+
 appIcon.setToolTip('Google Play Music');
-appIcon.on('double-click', () => {
-  mainWindow.setSkipTaskbar(false);
-  mainWindow.show();
-});
+appIcon.on('click', toggleMainWindow);
+
+/* does not work if the 'click' handler is present */
+//appIcon.on('double-click', toggleMainWindow);
```
MarshallOfSound: Ummmmm, that is one hell of a bug.  Is that OS specific?  Which platform are you running?

src/main/features/core/tray.js
```diff
@@ -39,11 +39,66 @@ const setContextMenu = () => {
...
+      }
+    }
+  } else { // Windows, Linux
+    const wasMinimized = win.isMinimized();
+
+    if (wasMinimized || !win.isFocused()) {
+      win.setSkipTaskbar(false);
+      win.show();
+
+      win.focus(); // if not minimized, try to focus it
```
MarshallOfSound: On windows we minimize when we close to taskbar. in order to bring it back.  Use `win.restore()` or `win.show()`.  Both trigger a focus call

src/main/features/core/tray.js
MightyPork: Linux. Well, to be clear, it probably "does" work, but when you double-click it, it pops to foregorund and then goes back to the tray.

MarshallOfSound: Oh, single click goes twice :/

Strange

MightyPork:
So what remains to be resolved is the click / double-click.

src/main/features/core/tray.js
MightyPork: I've removed the whole focus thing, because it didn't work due to the other bug (can't get focus after closing the window with a cross button).

MightyPork:
@MarshallOfSound so what do you think of this?

I see you prefer the double-click, ideas:

- Make it configurable
- Single click on Linux only (maybe Mac? idk)

MarshallOfSound:
Consider this, one action needs to be opening the tray icon menu.  Another needs to be opening the app.

On windows thats like Right Click for context menu and double for show.
On mac double to show isn't a thing, and it is left click for context menu
On linux I'm assuming it's similar to mac?

MightyPork:
On Linux it should be

- left click to show / hide
- right click to show context menu

okay I'll make it the way you said, want to have this finally marged.

MightyPork:
Here we go.

Windows | Linux | Mac
---|---|---
Double left click to toggle | Left click to toggle | N/A, use the context menu

I hope I understood it right. Works fine on Linux / xfce, that's what I can test.

PR #309, opened 2016-03-07
Hide 'Send Gift' tab

tf2manu994:
Hides the Send Gift tab

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/309)


MarshallOfSound:
Seems legit

PR #310, opened 2016-03-07
Tempfix hiding account info

tf2manu994:
white screen with unreadable text. understandable if you dont want to merge.

Also, haven't had a chance to test this, coding on mobile is hard, might want to do that

sidenote: this actually has the send gift thing also removed, so this merges 2 things in one! in theory

i should learn how git works

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/310)

### DIFF
--- a/src/inject/GPMInject/interface/customUI.js
+++ b/src/inject/GPMInject/interface/customUI.js
@@ -35,13 +35,13 @@ window.wait(() => {
   hide('[data-action="upload-music"]');
   hide('[data-action="help-and-feedback"]');
   hide('[data-action="send-gift"]');
-  
-  //Temp Fix: Hiding the entire account details bubble
+
+  // Temp Fix: Hiding the entire account details bubble
   hide('[aria-label="Account Information"]');
   hide('[class="gb_ab"]');
   hide('[class="gb_bb"]');
 
-  //Hide Upload Dialog
+  // Hide Upload Dialog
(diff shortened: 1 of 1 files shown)


tf2manu994:
okay lets try now

tf2manu994:
are you serious

the problem was a stupid space before a comment
really

MarshallOfSound:
@tf2manu994 If you use atom you can use the `linter-eslint` plugin which will tell you as you edit the code the linting mistakes you make :+1:

src/inject/GPMInject/interface/customUI.js
```diff
@@ -34,7 +34,14 @@ window.wait(() => {
   // Settings options that won't work
   hide('[data-action="upload-music"]');
   hide('[data-action="help-and-feedback"]');
+  hide('[data-action="send-gift"]');
```
MarshallOfSound: This is also in another PR (I think yours o_0)

Be careful not to bleed commits / delta into other PR's / branches

MarshallOfSound:
Instead of hiding the account panel entirely, we should fix the underlying issue of the user not being able to read the text.  This appears to be a result of the theming library.

tf2manu994:
ah okay

see the send gift hiding bleeding into another pull request was a good thing hehe

i should learn git

PR #316, opened 2016-03-07
added 2 spaces to make a line

tf2manu994:
[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/316)

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ We run tests and generate signed installers on three CI platforms
 * Linux --> [Circle CI][4]
 
 You can download the latest signed installers for Windows from the artifacts section of AppVeyor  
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
These small PR's are becoming annoying (I like the enthusiasm) but can we try make one larger PR with lots of small grammar / styling fixes

:+1:

tf2manu994:
@MarshallOfSound i would but these 2 fixes are on different branches, master and website

MarshallOfSound:
I know, I'm just saying please make sure that there isn't going to be "just one more" after this :+1: 

Go through a thourough check now if this is what you are doing

PR #343, opened 2016-03-07
Fix #306 and restore original color on profile popup buttons

### LINKED ISSUE #306: Clicking on avatar brings up  thing that isn't themed and has white text on a white background

MightyPork:
Ideally the theme should take care of this, but for now *grey on white* is better than *white on white*.

xxx

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/343)

### DIFF
--- a/src/inject/GPMInject/interface/customUI.js
+++ b/src/inject/GPMInject/interface/customUI.js
@@ -21,11 +21,11 @@ const style = (elementSelector, styleObject) => {
   });
 };
 
-const cssRule = (css) => {
-  const style = document.createElement('style');
-  style.type = 'text/css';
-  style.appendChild(document.createTextNode(css));
-  document.head.appendChild(style);
+const cssRule = (styles) => {
+  const tag = document.createElement('style');
+  tag.type = 'text/css';
+  tag.appendChild(document.createTextNode(styles));
+  document.head.appendChild(tag);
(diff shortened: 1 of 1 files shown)


src/inject/GPMInject/interface/customUI.js
```diff
@@ -21,7 +21,17 @@ const style = (elementSelector, styleObject) => {
...
+const cssRule = (styles) => {
+  const tag = document.createElement('style');
+  tag.type = 'text/css';
+  tag.appendChild(document.createTextNode(styles));
+  document.head.appendChild(tag);
+};
+
 window.wait(() => {
+  // Restore original text color on buttons in the Profile dialog (dark theme sets it to white).
+  cssRule('.gmusic-theme .material div.gb_ob a.gb_Ca.gb_nb {color: #666 !important}');
```
MarshallOfSound: Just going to point out here, that those classes you used won't work cross language/locale.  For some reason those classes change when going between different countries and using Google Play music.

Also this really should be PR'ed into https://github.com/gmusic-utils/gmusic-theme.js instead of patched here

MarshallOfSound:
Closing in favor of a PR into https://github.com/gmusic-utils/gmusic-theme.js

Issue #217, opened 2016-03-05
Cannot go forwards or back between pages on Windows and OSX Version


joslak:
If you click onto another section of GPM you cannot go backwards using either a back button or macro on a mouse. Not sure if a bug or missing feature but definitely affects usability of program. Confirmed to be issue on OSX and Windows version of program.

DougieNeish:
Just wanted to second this feature request.

flakeparadigm:
This seems to be a design choice made for v3. #198

jheiselman:
Man, I wish GitHub would implement the star system that Google has for issues so I can voice my interest in this issue without having to comment.

Anyway, definitely going to follow this one as it's the only one huge issue for usability for me.

mstrelan:
Menu > Settings > Add your music. This gets you stuck on "Add your existing music collection to Google Play". Clicking "Install" does nothing, and there are no links or buttons to go back. Can't press backspace, or alt+left or any other "back" shortcut. Have to kill the app or right click on the systray icon to close it.

eXtreme:
Well for me it is also a step backwards (in overall a great piece of software). 

Saying that: 

>  if you can't use a website without forward/back navigation buttons the website is doing something wrong. 

is IMHO a little bit wrong, using websites with back/forward buttons is a natural thing to do and doing that in browser is always possible (well, when page changes url, history works). And with such UI design which GPM has, having a back button is a must. On android you have a hardware button for that. On iOS app designers have to include back button in UI. The screen on mobile devices is small, you cannot add more parent "context" to the view because there is no space. With large workspace like webpage or desktop app like this, with material design there is loads of wasted space.

The problem with GPM UI is that the whole view changes with no context of previous view. Consider searching a playlist, then clicking on an artist (to see more his/her songs). You can't easily get back to that playlist, you have to search it again (and good luck if it wasn't like one of the first, and all playlists have almost the same name). There are many more examples where back button is helpful. 
Spotify desktop app has back/forward button as well.

MarshallOfSound:
OK guys, maybe on this one I started out with the wrong opinion.  Idk...  But clearly a lot of people rely on those navigation arrows so I'll look at implementing them.

*Falls to peer pressure of github issues*

joslak:
[...]

PR #348, opened 2016-03-07
Added an injected Back button w/ auto-hide

MightyPork:
This solves #217 and it's dupes.

I've added a back button, forward button doesn't seem to be necessary (I've never needed it, anyway).

The button auto-hides on the "Listen Now" page, which is the initial landing page.

screenshot_2016-03-07_21-00-09

screenshot_2016-03-07_21-00-42

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/348)

### DIFF
--- a/src/inject/GPMInject/interface/customUI.js
+++ b/src/inject/GPMInject/interface/customUI.js
@@ -21,6 +21,13 @@ const style = (elementSelector, styleObject) => {
   });
 };
 
+const cssRule = (styles) => {
+  const tag = document.createElement('style');
+  tag.type = 'text/css';
+  tag.appendChild(document.createTextNode(styles));
+  document.head.appendChild(tag);
+};
+
 window.wait(() => {
   // Top left account control buttons
   hide('#material-one-right #gb > div > div > div:not(:last-child)');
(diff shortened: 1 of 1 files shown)


MightyPork:
I didn't realize this will interfere with the "X" button when search term is entered.

So still something to be done...

MightyPork:
Okay the issue with the search clear button is now fixed!

**NOTE** this PR adds the same helper function `cssRule` as #343.
There shouldn't be any conflict, so just a heads-up. I didn't see other way to do it cleanly.

src/inject/GPMInject/interface/customUI.js
```diff
@@ -53,4 +60,40 @@ window.wait(() => {
     return false;
   });
   document.querySelectorAll('.nav-section.material')[0].insertBefore(dSettings, document.querySelectorAll('.nav-section.material > a')[2]); // eslint-disable-line
+
+  // Back button
+  const back = document.createElement('paper-icon-button');
+  back.setAttribute('onclick', 'history.back()');
```
MarshallOfSound: Can we change this to the more syntactically correct

```js
back.addEventListener('click', () => history.back());
```

MightyPork: sure thing, done

MarshallOfSound:
I'm going to merge this once CI finishes and test from there.  I already have too many remotes on my git client :+1:

MightyPork:
it's finally completed. those things take ages...

juschu:
Nice to see a nicer version of a back button.
But besides the interference with the X, I think the search field is a bad place for that button anyway. It's not self explaining and the users will think it has something to do with the search, which it does not.
I think a <button> outside of the search field would be better.

MightyPork:
Sure, that's a thing I considered too, might be better. Maybe on the left from the search field?
Feel free to improve it ;-)

PR #352, opened 2016-03-07
Mini player tweaks

### LINKED ISSUE #351: Cannot enlarge Mini-player
I am trying to control my window sizes and locations, and would like the size of the Mini-player to be a little bigger, 350x350, but have found that I cannot do so.

I suspect it is due to the following hint that I found when running xprop on the mini-player:
```
WM_NORMAL_HINTS(WM_SIZE_HINTS):
		program specified location: 1892, 169
		program specified maximum size: 310 by 310
```

I believe this should probably be setting the size (Preferred) or minimum size, not the max size, to allow for resizing of the mini player.

Reference: https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472494112
https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.2.3
> The WM_SIZE_HINTS.flags bit definitions are as follows:
> Name	Value	Field
> ...
> PSize	8	Program-specified size

Additionally, it would be good to have some attribute, such as the Window title, to change when in Mini player mode, for easier controlling of the window by automated tools.

**OS:** Linux (Arch)

**Issue Description:** Mini-window cannot be resized greater than initial 310x310

**Steps to Reproduce:**
1. Start GPMDP
2. Send to Mini-Player mode
3. Attempt to resize window
4. Find that it cannot be bigger, but can be smaller than 310x310

StephenBrown2:
These changes fix #351 for me.

- Remove MaximumSize restriction on MiniPlayer (so it can be greater than 310x310)
- Change title when MiniPlayer mode is engaged (so it can be affected differently by Window manager tools)

### DIFF
--- a/src/inject/GPMInject/interface/mini.js
+++ b/src/inject/GPMInject/interface/mini.js
@@ -16,7 +16,6 @@ window.wait(() => {
     Emitter.fire('mini', { state: true });
     oldSize = remote.getCurrentWindow().getSize();
     mainWindow.setSize(MINI_SIZE, MINI_SIZE);
-    mainWindow.setMaximumSize(MINI_SIZE, MINI_SIZE);
     webContents.executeJavaScript('document.body.setAttribute("mini", "mini")');
     remote.getCurrentWebContents().setZoomFactor(1);
     remote.getCurrentWindow().setAlwaysOnTop(Settings.get('miniAlwaysOnTop', false));
--- a/src/inject/GPMInject/interface/mini.js
+++ b/src/inject/GPMInject/interface/mini.js
@@ -16,6 +16,7 @@ window.wait(() => {
     Emitter.fire('mini', { state: true });
     oldSize = remote.getCurrentWindow().getSize();
     mainWindow.setSize(MINI_SIZE, MINI_SIZE);
+    mainWindow.setTitle('GPMDP - Mini Player');
     webContents.executeJavaScript('document.body.setAttribute("mini", "mini")');
     remote.getCurrentWebContents().setZoomFactor(1);
     remote.getCurrentWindow().setAlwaysOnTop(Settings.get('miniAlwaysOnTop', false));
@@ -29,6 +30,7 @@ window.wait(() => {
     //        --> https://github.com/atom/electron/blob/master/atom/browser/native_window_views.cc
     mainWindow.setMaximumSize(99999999, 999999999);
     mainWindow.setSize(...oldSize);
+    mainWindow.setTitle('Google Play Music Desktop Player');
     webContents.executeJavaScript('document.body.removeAttribute("mini", "mini")');
     remote.getCurrentWebContents().setZoomFactor(1);
     remote.getCurrentWindow().setAlwaysOnTop(false);


src/inject/GPMInject/interface/mini.js
```diff
@@ -16,7 +16,7 @@ window.wait(() => {
     Emitter.fire('mini', { state: true });
     oldSize = remote.getCurrentWindow().getSize();
     mainWindow.setSize(MINI_SIZE, MINI_SIZE);
-    mainWindow.setMaximumSize(MINI_SIZE, MINI_SIZE);
```
MarshallOfSound: I'm not sure removing this is the best idea.  Some users will end up resizing to be 400,500 or even 600x600 and then it will look awful because the album art is only 300x300

StephenBrown2: Perhaps .setAspectRatio(1) could be used? Restricting the maximum size doesn't prevent the issue of resizing smaller at say 200x250, and I would like it to be flexible enough to work with different people's workflows (namely, mine :-p).

IMO, the album art may be a 300x300 image, but it is not guaranteed to be "crisp and clear" at that resolution. Source: tested resizing larger and smaller and image quality doesn't seem to change when it's not it's native 300^2.

MarshallOfSound: `setAspectRatio(1)` only works on OSX as far as I am aware.  The issue isn't 1-1 ratio, I have enforced that in a resize handler.  The issue is people resizing larger than the album artwork.  The artwork will just stretch and look awful.

My question I guess is what are you trying to achieve by making the mini player not so mini :+1:

StephenBrown2: Ah good, I did notice that it held to a square ratio, but I'd assumed it was due to me adding setAspectRatio.

I suppose our opinions of "awful" vary as well. At 350px it still looks darn spiffy to me. ;-) But I can see it not looking as good at, say 500 or 750 or 1000px. But at that point, you're right, it's no longer "mini".

Here's my current flow, with my chat and music on one monitor in a semi-tiled manner:
workspace 1_014
Previously I had the Google Play Music Extension mini-window in the place where GPMDP is now, and it resized fine to fit in that space.

In consideration, 40 pixels is a really silly thing to debate, so I'm alright with adjusting my screen real-estate accordingly.
workspace 1_015
If you find a way to get actual hi-def album art from GPM, then perhaps the restriction can be lifted, but for now, I'll retract my request for expandability and can revert this commit.

The title change though, is much more crucial to my workflow, being able to position or float the window based on it's title, because it does serve quite a different purpose in full mode vs mini mode.

MarshallOfSound: I really like that layout, especially the pixelated font :laughing: 

But in all seriousness I get where you are coming from with the whole "its 40 pixels" thing.  How about a compromise, there was another PR a while back where we added a setting option but did not add a UI for it.

What if we added a hidden option for maximum mini player size?

StephenBrown2: Hehe, sorry for the radio silence. Don't want to be sharing company secrets from my work machine! (Nothing terribly important though. ;-p)

I'm all about options, it'd be good for those of us that "need" or would like it to be adjustable.

Just curious, is #390 the fix for tiling issues mentioned in #351 ?

MarshallOfSound: @StephenBrown2 I believe that is the fix yes

berryelectronics: @MarshallOfSound Doesnt GPM uses a bigger album art for that fullscreen album cover thing?
[Heres an example](https://lh3.googleusercontent.com/1CAgnjsHDEsBXCryizRO5b3Am6xwMlzEk3bED_HdMfk8HyF598nG5SyZlUHpjD-owpEJPM0nBw=s1280-c-e100-nu)

MarshallOfSound:
This is beyond out of date and I stand by my previous statements.  Closing this PR for now :+1:

Issue #298, opened 2016-03-07
Media Keys in Linux -> MPRIS Implementation


ZenHarbinger:
Running Ubuntu 16.04 64-bit, the media keys to not play/pause, next, or previous for me.

MarshallOfSound:
Not a clue what could be going on here, I'll let a more linuxy person come up with some ideas :P

parkerlreed:
Media keys can be implemented via MPRIS https://specifications.freedesktop.org/mpris-spec/latest/

This allows the player to publish metadata and be controlled from a central system.

MarshallOfSound:
That's on a post-it note somewhere on my desk, there's a node module for it that didn't seem to hard to implement :)

ZenHarbinger:
I'll try to take a look at it when I get a chance.  I've done a little bit w/ audio players and Ubuntu years ago; but this is magnitudes larger than anything I've done before.

MarshallOfSound:
@ZenHarbinger Sounds great, anything you can do to help would be much appreciated :+1: 

Just imagine how I feel, a couple of years ago I was puzzling of year 11 algebra.....

pszi1ard:
Media keys do  not work for me either on Ubuntu 14.04 (are they supposed to work at all?); the custom shortcuts do work, though.

parkerlreed:
@pszi1ard It's only a pull request at the moment. It hasn't been merged.

jostrander:
@pszi1ard Can you try doing this:

http://ubuntuforums.org/showthread.php?t=2241639&p=13110153#post13110153

and see if it helps?

pszi1ard:
@jostrander Sure, it does, but it disables it for all other applications, so unfortunately it's not a solution that I can use.

PR #368, opened 2016-03-08
MPRIS Integration

jostrander:
Just dropping this in as a start to mpris (#298) for anyone to continue working on or test out. It works using `dbus-send` on my system and currently broadcasts on `org.mpris.MediaPlayer2.google-play-music-desktop-player` with title, artist, album, albumArt, length, etc.

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/368)

### DIFF
--- a/src/main/features/linux/mprisService.js
+++ b/src/main/features/linux/mprisService.js
@@ -1,0 +1,84 @@
+import { app } from 'electron';
+import mpris from 'mpris-service';
+import _ from 'underscore';
+
+function mprisService() {
+  const mainWindow = WindowManager.getAll('main')[0];
+  let _songInfo = {};
+
+  const player = mpris({
+    name: 'google-play-music-desktop-player',
+    identity: 'Google Play Music Desktop Player',
+    canRaise: true,
+    supportedInterfaces: ['player'],
+    desktopEntry: 'google-play-music-desktop-player',
+  });
+  player.playbackStatus = 'Stopped';
+
+  player.on('raise', () => {
+    mainWindow.setSkipTaskbar(false);
+    mainWindow.show();
+  });
+
+  player.on('quit', () => {
+    app.quit();
+  });
+
+  player.on('play', () => {
+    if (!PlaybackAPI.isPlaying()) {
+      Emitter.sendToGooglePlayMusic('playback:playPause');
+    }
+  });
+
+  player.on('playpause', () => {
+    Emitter.sendToGooglePlayMusic('playback:playPause');
+  });
+
+  player.on('next', () => {
+    Emitter.sendToGooglePlayMusic('playback:nextTrack');
+  });
+
+  player.on('previous', () => {
+    Emitter.sendToGooglePlayMusic('playback:previousTrack');
+  });
+
+  player.on('stop', () => {
+    Emitter.sendToGooglePlayMusic('playback:stop');
+    player.playbackStatus = 'Stopped';
+  });
+
+  PlaybackAPI.on('change:song', (newSong) => {
+    player.metadata = _songInfo = {
+      'mpris:artUrl': newSong.albumArt,
+      'xesam:asText': (newSong.lyrics || ''),
+      'xesam:title': newSong.title,
+      'xesam:album': newSong.album,
+      'xesam:artist': newSong.artist,
+    };
+  });
+
+  PlaybackAPI.on('change:time', (time) => {
+    const newPosition = time.current * 1e3;
+    const newTotal = time.total * 1e3;
+    const delta = newPosition - player.position;
+
+    if (_songInfo && _songInfo['mpris:length'] !== newTotal) {
+      player.metadata = _.extend(_songInfo, {
+        'mpris:length': newTotal,
+      });
+    }
+
+    if (Math.abs(delta) > 2e6) {
+      player.seeked(delta);
+    } else {
+      player.position = newPosition;
+    }
+  });
+
+  PlaybackAPI.on('change:state', (playbackState) => {
+    // DEV: We skip stopped here because PlaybackAPI only shuttles true/false from GPM
+    player.playbackStatus = (playbackState) ? 'Playing' : 'Paused';
+  });
+}
+
+mprisService();
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
   },
   "optionalDependencies": {
     "dbus": "MarshallOfSound/node-dbus#linux-only",
+    "mpris-service": "^1.1.1",
     "electron-installer-debian": "^0.2.0",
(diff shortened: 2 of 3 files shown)


src/main/features/linux/mprisService.js
```diff
@@ -0,0 +1,60 @@
+const mpris = require('mpris-service');
```
MarshallOfSound: Please use the `import` syntax

```js
import mpris from 'mpris-service';
```

It's just the ES6 way :+1:

src/main/features/linux/mprisService.js
```diff
@@ -0,0 +1,60 @@
...
+        this.player = mpris({
+            name: 'gpmdp',
+            identity: 'Google Play Music Desktop Player',
+            supportedInterfaces: ['player']
+        });
+
+        this.listeners();
+    }
+
+    listeners() {
```
MarshallOfSound: More of a style than anything else but "private" methods I.e. methods that are only called internally.  Can they start with an `_` :+1: 

So this would be `_listeners()`

src/main/features/linux/mprisService.js
```diff
@@ -0,0 +1,60 @@
...
+        PlaybackAPI.on('change:song', (newSong) => {
+            this.updateMetadata(newSong);             
+        });
+    }
+
+    updateMetadata(newSong) {
+        this.player.metadata = {
+            // This is required in MPRIS, but no uuid for each track is given by gmusic library.
+            //'mpris:trackid': player.objectPath('track/0'),
+            //'mpris:length' : newSong.length, //
```
MarshallOfSound: For song length playbackAPI has a `on('time')` I think or something similar

src/main/features/linux/mprisService.js
```diff
@@ -0,0 +1,60 @@
...
+
+        PlaybackAPI.on('change:song', (newSong) => {
+            this.updateMetadata(newSong);             
+        });
+    }
+
+    updateMetadata(newSong) {
+        this.player.metadata = {
+            // This is required in MPRIS, but no uuid for each track is given by gmusic library.
+            //'mpris:trackid': player.objectPath('track/0'),
```
MarshallOfSound: Hmmmm, can we use the `last.fm` service to pull a UUID or can this just be any old UUID?

Something like `Date.now()`?

MarshallOfSound:
Woo hoo, cool things are happening.

But the CI is failed :cry:

jostrander:
Another issue I noticed with CI, this now requires node-dbus, but node-dbus is a linux specific package. Any ideas on how to only require it for linux?

jostrander:
whoops, wrong button

MarshallOfSound:
See the `optionalDependencies` section of the `package.json`

Anything in there will only be installed if it can be, but won't throw an error if it can't :+1: 

Add the entire `mpris` package into that section

src/main/features/linux/mprisService.js
jostrander: Honestly I'm not sure it's entirely required, after reviewing docs here:

https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#index1h4

It seems to be relevant when using TrackList only, which I didn't implement because gmusic doesn't implement playlist features.

src/main/features/linux/mprisService.js
jostrander: What unit of seconds is that using? mpris is expecting micro seconds.

jostrander:
Also ci/circle CI is failing because of eslint, it hates me. Is there a CS tool I can use to clean that file up?

MarshallOfSound:
If you use atom there is an eslint plugin that I use.  Highlights things red as you type them to show you that something is wrong.

MarshallOfSound:
@jostrander Might be worth giving this a once over

https://github.com/twolfson/google-music-electron/blob/master/lib/mpris.js

jostrander:
I took a look at the code there, I implemented the seek and playerStatus, however the desktopEntry is what seems to be needed in order for the Ubuntu sound menu to correctly work. Unfortunately it seems the mpris-service library hasn't implemented something correctly because asking for it's desktopEntry returns "Not Supported". Making manual changes to the mpris-service currently to see if its something I can fix upstream.

jostrander:
I do want to point out, even with MPRIS being 90% implemented, media keys still do not work on my system. There seems to be a work around for the spotify client in linux as well that uses XBindKeys and MPRIS:

https://wiki.archlinux.org/index.php/spotify#D-Bus

jostrander:
@MarshallOfSound You want to test this out and see if there's anything I've missed. It seems to be working for all features except volume level :+1:. Just going to leave that to the OS to handle. I've removed any mention of media keys because that's not something listed in the spec for mpris unfortunately. :-1:

jostrander:
Resolved merge conflicts from gnome media keys.

src/main/features/linux/mprisService.js
```diff
@@ -0,0 +1,84 @@
...
+  });
+  player.playbackStatus = 'Stopped';
+
+  player.on('raise', () => {
+    mainWindow.setSkipTaskbar(false);
+    mainWindow.show();
+  });
+
+  player.on('quit', () => {
+    app.quit();
```
MarshallOfSound: Something to think about here, should this act as a quit, even if we have minimize to task bar turned on?

(I'm not sure how this quit command would be received)

jostrander: This quit should be received only if the mpris client sends a quit, generally speaking the client might want to switch to a different media source, so quitting should in fact quit. The spec does say that this should cause the client to close, but doesn't have to be implemented.

MarshallOfSound: Fair enough :+1:

MarshallOfSound:
Sorry @jostrander I forgot where we got to on this.  Is it good to go?

jostrander:
Yes this is good to go, just have to fix the merge conflict. The only thing I notice could be an issue is building with node 4.4.0 and needing to use electron-rebuild because this uses dbus as well indirectly.

MarshallOfSound:
Ahh never pushed that Linux electron rebuild script thing.  It's sitting in a VM somewhere.  I'll push that when I get home from holiday

jostrander:
Well, rebased and fixed the conflict, which was the dbus change from package.json, which might change again when you push your rebuild script, but if that happens tag me and I'll update again.

mcg:
Seems like this is broken again(at least media keys in 16.04) since 3.2.5?

MarshallOfSound:
I'd it worked in 3.2.4 (it did) it should work in 3.2.5

MarshallOfSound:
Nothing changed

PR #386, opened 2016-03-08
Settings toggle for native window decorations

MightyPork:
Yep, it's me again, with another PR that will no doubt fail thanks to Travis.

- The window decorations are now configurable.
- By default nothing should change, but there's a toggle in the settings for the custom frame.
- The settings window is slightly higher (yet again), it didn't fit

screenshot_2016-03-08_20-33-32

[Review on Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/386)

### DIFF
--- a/src/assets/less/main_window.less
+++ b/src/assets/less/main_window.less
@@ -13,7 +13,7 @@ html, body {
 }
 
 // Only use the custom borders on windows
-.darwin, .linux, .freebsd, .sunos {
+.native-frame {
   .window-border {
     border: 0;
     > header {
--- a/src/assets/less/main_window.less
+++ b/src/assets/less/main_window.less
@@ -12,7 +12,7 @@ html, body {
   padding: 0;
 }
 
-// Only use the custom borders on windows
+// Hide custom frame when native frame is used
 .native-frame {
   .window-border {
     border: 0;
--- a/src/main/utils/Settings.js
+++ b/src/main/utils/Settings.js
@@ -45,7 +45,7 @@ class Settings {
 
   _load() {
     const userSettings = JSON.parse(fs.readFileSync(this.PATH, 'utf8'));
-    this.data = _.defaults(userSettings, initalSettings);
+    this.data = _.extend({}, initalSettings, userSettings);
   }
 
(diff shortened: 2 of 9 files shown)


src/public_html/desktop_settings.html
```diff
@@ -75,6 +75,10 @@ <h5 class="theme-text">Choose your theme highlight color</h5>
                 <input type="checkbox" class="filled-in" id="playback-api" />
                 <label for="playback-api">Enable Playback API <i>This allows other applications to interface with GPMDP</i></label>
               </p>
+              <p>
+                <input type="checkbox" class="filled-in" id="native-frame" />
+                <label for="native-frame">Use system window decorations <i>Requires restart.</i></label>
```
MarshallOfSound: Can we change this to `Use system window border styles`.  The average user might understand that a bit more?

src/main/utils/Settings.js
```diff
@@ -43,7 +44,8 @@ class Settings {
   }
 
   _load() {
-    this.data = JSON.parse(fs.readFileSync(this.PATH, 'utf8'));
+    const userSettings = JSON.parse(fs.readFileSync(this.PATH, 'utf8'));
+    this.data = _.defaults(userSettings, initalSettings);
```
MarshallOfSound: I like to avoid object mutation at all costs, can we change this to

```js
_.extend({}, initialSettings, userSettings);
```

src/assets/less/main_window.less
```diff
@@ -13,7 +13,7 @@ html, body {
 }
 
 // Only use the custom borders on windows
```
MarshallOfSound: This comment is somewhat inaccurate now

src/public_html/desktop_settings.html
MightyPork: That's not any more clear than what I have there now.
I'd go with "Use system window borders" then.

MightyPork:
@MarshallOfSound should be all good now. Can you also pls have a look at #297 when you're at it, if it can be merged already?

Issue #270, opened 2016-03-06
Some issues with i3 tiling WM.


kminehart:
I'm not using a DE, only the WM called "i3".  See here:  https://i3wm.org/

I think the issues might be because of the custom window borders and not allowing the WM to control the application?  For example, I can't super+shift+q to exit, which executes `kill`.  Clicking the X, fullscreen, or minimize does nothing either.

I'll see if I can do anything to help with this, I made a fork for my changes:

https://github.com/pascalsoftware/Google-Play-Music-Desktop-Player-UNOFFICIAL-

kminehart:
More specifically, the app is ignoring or SIGKILL.

AnotherKamila:
Also seeing this in i3.

Additionally, if I switch to the miniplayer while in tiling mode, the window size is different than what the miniplayer expects, and it becomes very unhappy about this (constantly redrawing itself to the expected and actual size in a loop). As the miniplayer expects a fixed window size, probably telling the WM to float it is the most reasonable behavior (or you could make it resizable => me happy :D).

kminehart:
Update: 
in `src/index.js` I added 
```
    // Emitted when the user tries to close the application.
    mainWindow.on('close', () => {
      console.log('close');
    });
```
and when I make an attempt to close the application, either with super+shift+q, or with the x, or ctrl+w, the event gets fired and 'close' is printed on the console.  At this point I'm not familiar enough with the application to correctly handle a close, but I'm working on it.

kminehart:
Here's the perpetrator:

So at the moment if anyone else is having this problem, make sure to disable "minimize to tray" in the desktop settings, then it should work fine!  It looks like i3 and electron don't like to play nice when trying to hide a window, since i3 really has no way to hide a window.

```JavaScript
mainWindow.on('close', (event) => {
  if ((Settings.get('minToTray', true) || process.platform === 'darwin') && !global.quiting) {
    if (process.platform !== 'darwin') {
      mainWindow.minimize();
      mainWindow.setSkipTaskbar(true);
    } else {
      if (PlaybackAPI.isPlaying()) {
        Emitter.sendToGooglePlayMusic('playback:playPause');
      }
      mainWindow.hide();
    }
    event.preventDefault();
    return;
  }
```

AnotherKamila:
[...]

PR #390, opened 2016-03-09
Fixed mini-player on i3 (Linux)

N0ps32:
I also ran into the issue discussed in this thread: https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/issues/270

This pull request makes sure the mini-player works in i3 by using the i3 IPC bridge and sending the window into floating mode.

* [x] Fix linting issues
* [x] Add global flag to RegExp

### DIFF
--- a/src/inject/GPMInject/interface/mini.js
+++ b/src/inject/GPMInject/interface/mini.js
@@ -7,12 +7,15 @@ const MINI_SIZE = 310;
 let mini = false;
 let oldSize;
 
+const I3IpcHelper = remote.getGlobal('I3IpcHelper');
+
 window.wait(() => {
   if (Settings.get('miniAlwaysShowSongInfo', false)) {
     document.body.setAttribute('controls', 'controls');
   }
 
   window.GPM.mini.on('enable', () => {
+    I3IpcHelper.setFloating(true);
     Emitter.fire('mini', { state: true });
     oldSize = remote.getCurrentWindow().getSize();
     mainWindow.setSize(MINI_SIZE, MINI_SIZE);
@@ -24,6 +27,7 @@ window.wait(() => {
   });
 
   window.GPM.mini.on('disable', () => {
+    I3IpcHelper.setFloating(false);
     Emitter.fire('mini', { state: false });
     // DEV: Set max size to be massive
     //      Same reason as specified in Electron src
--- a/src/inject/GPMInject/interface/mini.js
+++ b/src/inject/GPMInject/interface/mini.js
@@ -7,15 +7,12 @@ const MINI_SIZE = 310;
 let mini = false;
 let oldSize;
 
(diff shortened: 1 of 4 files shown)


MarshallOfSound:
Linting errors, please run `npm test` before raising a PR :+1: It saves the build agents some grunt work

src/main/utils/I3IpcHelper.js
```diff
@@ -0,0 +1,40 @@
...
+    return false;
+  }
+
+  _getWindowTitleRegex() {
+    const title = WindowManager.get(global.mainWindowID).getTitle();
+    return '^' + title + '$';
+  }
+
+  _escapeShellArg(str) {
+    return str.replace(/'/, "\\'").replace(/"/, "\\\"");
```
MarshallOfSound: Do these regexp's need the `/g` flag for global searching?

src/main/utils/I3IpcHelper.js
```diff
@@ -0,0 +1,40 @@
...
+
+class I3IpcHelper {
+
+  isI3() {
+    return WindowManager.getWindowManagerName() === 'i3';
+  }
+
+  _hasI3Msg(callback) {
+    if(this.isI3()) {
+      exec('command -v i3-msg', error => {
```
MarshallOfSound: Pretty sure error needs brackets around it to meet the linting spec, if not linting then consistency :+1:

src/main/utils/WindowManager.js
```diff
@@ -65,6 +65,14 @@ class WindowManager {
       this.windows[windowID].close();
     }
   }
+
+  getWindowManagerName() {
+    if(process.platform === 'linux') {
+      return process.env.XDG_CURRENT_DESKTOP;
+    }
+    return 'unknown';
```
MarshallOfSound: Might make more syntactical sense to `return undefined`;

N0ps32:
The problems should be fixed now! :+1: 
Maybe you should mention that in the documentation. I haven't found anything regarding tests/linting in the readme when I looked yesterday.

src/inject/GPMInject/interface/mini.js
```diff
@@ -7,12 +7,15 @@ const MINI_SIZE = 310;
 let mini = false;
 let oldSize;
 
+const I3IpcHelper = remote.getGlobal('I3IpcHelper');
```
MarshallOfSound: Is there really no better way to do this than using `getGlobal`.   For some reason it seems a hacky workaround :cry: 

I'm still probably going to merge it but instead of using `getGlobal` is it possible to use the `Emitter` classes?  Or is that not going to work for some reason?

N0ps32:
I use events instead of globals now and set-up the event listeners in index.js.
Is there a better place to put those 2 lines?

PR #394, opened 2016-03-09
Use electron's `userData` for config files

### LINKED ISSUE #267: Linux config folder change
The config folder on Linux should really be `~/.GPMDP_STORE` or `~/.config/GPMDP_STORE` instead of `~/GPMDP_STORE`. That way there wouldn't be an extra visible folder in the user's home directory, which many people (including me) don't like.

jacobmischka:
If a file isn't present in `userData`, look for it in its old spot. If
it's in its old spot, move it to `userData` and try to delete the old
directory.

Fixes #267. Closes #287.

### DIFF
--- a/src/main/utils/PlaybackAPI.js
+++ b/src/main/utils/PlaybackAPI.js
@@ -2,19 +2,24 @@ import fs from 'fs';
 import mkdirp from 'mkdirp';
 
 const os = require('os');
+const app = require('app');
 
 const environment = process.env;
 
 class PlaybackAPI {
   constructor() {
-    const DIR = (environment.APPDATA ||
+    const DIR = app.getPath('userData');
+    this.PATH = `${DIR}/playback.json`;
+    const OLD_DIR = (environment.APPDATA ||
       (process.platform === 'darwin' ? environment.HOME + '/Library/Preferences' : os.homedir())) +
       '/GPMDP_STORE';
-    this.PATH = `${DIR}/playback.json`;
+    const OLD_PATH = `${OLD_DIR}/playback.json`;
 
     this.reset();
-
-    if (!fs.existsSync(this.PATH)) {
+    if (fs.existsSync(OLD_PATH)) {
+      fs.renameSync(OLD_PATH, this.PATH);
+      fs.rmdir(OLD_DIR, function () {});
+    } else if (!fs.existsSync(this.PATH)) {
       mkdirp(DIR);
       this._save();
     }
--- a/src/main/utils/PlaybackAPI.js
+++ b/src/main/utils/PlaybackAPI.js
@@ -1,4 +1,4 @@
-import app from 'app';
+import { app } from 'electron';
 import fs from 'fs';
 import mkdirp from 'mkdirp';
 
--- a/src/main/utils/PlaybackAPI.js
+++ b/src/main/utils/PlaybackAPI.js
@@ -1,8 +1,8 @@
+import app from 'app';
 import fs from 'fs';
 import mkdirp from 'mkdirp';
 
 const os = require('os');
-const app = require('app');
 
 const environment = process.env;
 
--- a/src/main/utils/Settings.js
+++ b/src/main/utils/Settings.js
@@ -1,9 +1,9 @@
+import app from 'app';
 import fs from 'fs';
 import mkdirp from 'mkdirp';
 import initalSettings from './initialSettings';
 
 const os = require('os');
-const app = require('app');
 
 const environment = process.env;
 
--- a/src/main/utils/Settings.js
+++ b/src/main/utils/Settings.js
@@ -1,4 +1,4 @@
-import app from 'app';
+import { app } from 'electron';
 import fs from 'fs';
 import mkdirp from 'mkdirp';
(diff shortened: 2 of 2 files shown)


jacobmischka:
The os one was there already, unsure if it's in within the scope of this pull request/issue to make that change. I'd be happy to change it if it's agreed that it is.

jostrander:
Agreed, it's ultimately up to @MarshallOfSound, I just know he requested I do the same thing. I did notice the OS one was required in, not sure that it matters.

MarshallOfSound:
@jacobmischka Just implement your new one as an `import` and I'll fix up the dodgy `require` straight to master.  Not sure how that one slipped by my :cactus:

jacobmischka:
Oh, okay. Saw that a second after I pushed, give me one sec to undo the os change.

jacobmischka:
Okay, should be good now.

src/main/utils/PlaybackAPI.js
```diff
@@ -1,3 +1,4 @@
+import app from 'app';
```
MarshallOfSound: This is the old style for electron it should be

```js
import { app } from 'electron';
```

src/main/utils/Settings.js
```diff
@@ -1,3 +1,4 @@
+import app from 'app';
```
MarshallOfSound: Same as other file

jacobmischka:
Oops, I haven't used ES6 imports at all yet until now, my mistake.

src/main/utils/PlaybackAPI.js
```diff
@@ -7,14 +8,18 @@ const environment = process.env;
...
       '/GPMDP_STORE';
-    this.PATH = `${DIR}/playback.json`;
+    const OLD_PATH = `${OLD_DIR}/playback.json`;
 
     this.reset();
-
-    if (!fs.existsSync(this.PATH)) {
+    if (fs.existsSync(OLD_PATH)) {
+      fs.renameSync(OLD_PATH, this.PATH);
+      fs.rmdir(OLD_DIR, function () {});
```
MarshallOfSound: In ES6 this line should be

```js
fs.rmdir(OLD_DIR, () => {});
```

No need for the "function" declaration

src/main/utils/Settings.js
```diff
@@ -8,15 +9,21 @@ const environment = process.env;
...
-    this.PATH = `${DIR}/${(jsonPrefix || '')}.settings.json`;
+    const OLD_PATH = `${OLD_DIR}/${(jsonPrefix || '')}.settings.json`;
     this.data = initalSettings;
     this.lastSync = 0;
 
     if (fs.existsSync(this.PATH) && !wipeOldData) {
       this._load();
+    } else if (fs.existsSync(OLD_PATH)) {
+      fs.renameSync(OLD_PATH, this.PATH);
+      fs.rmdir(OLD_DIR, function () {});
```
MarshallOfSound: Same as other comment

MarshallOfSound: Can you confirm your logic for the asyncronous usage of rmdir here?

I don't think it's a problem and will actually be better performance :+1: 

Was it deliberate?

MarshallOfSound:
@jacobmischka 2 small issues (basically 1) and a query and then it'll be good for merge :+1: :dancer:

MightyPork:
Why not put this to a separate file like it was in my PR here?
https://github.com/MightyPork/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/0d6cd0c6a745a5c1b23301d42b77357cb670fb93/src/main/utils/configDir.js

(also we should probably call it "gpmdp" on linux, no need for uppercase - and the _STORE doesn't make any sense)

MarshallOfSound:
I'm going to merge and do a quick fixup

MarshallOfSound:
@MightyPork I updated it in master to use centralized methods for creating the JSON file and handling old JSON files

src/main/utils/Settings.js
jacobmischka: Yeah, figured it doesn't matter when it's called or even if it succeeds or not

jacobmischka:
@MightyPork, it uses the app's name  when using userData. The other part is for backward compatibility, we can't rename that.

MightyPork:
ah okay, I didn't understand it right before.

Can confirm that it's working, settings are now with the rest of the browser stuff in `~/Google Play Music Desktop Player/json_store`. Seriously that's one dumb name for a folder, but now it must stay :(

jacobmischka:
The `Google Play Music Desktop Player` bit is because that's what the binary is set to be called. That's a different, larger issue to rename that. Though I agree, I'd also like that to be done.

image

MarshallOfSound:
@jacobmischka Yeah, we need to bite the bullet soon and rename the app.  Preferably to a one word name to elimate some SquirrelWindows issues

jacobmischka:
Honestly I don't really mind even `google-play-music-desktop-player`, it's just the caps and spaces that are problematic. `gpmdp` is probably fine too, just less verbose and potentially confusing.

PR #396, opened 2016-03-09
fix grammar

tf2manu994:
also, can someone show me a good guide on how to use git because i tried all the things on stackoverflow that were related to fixing merge problems and they all fell flat on their face


tf2manu994:
also, @MarshallOfSound, I decided I'll get ATOM and GitKraken, what's the linting plugin you suggest?

PR #427, opened 2016-03-10
Include Linux Path

N0ps32:


MarshallOfSound:
This is invalid as those paths are now completely incorrect.

My bad though, I should have updated the docs when I updated the code

Issue #401, opened 2016-03-09
Windows - mini player controls are off-screen


t-readyroc:
First noticed this last night after upgrading at home, & now I'm noticing it at work as well.  Once the player is minimized, controls appear off the side of the screen.

**OS:** - Windows 10 (both home & work)

**Issue Descriptions:** - mini player controls off-screen

**Steps to Reproduce:** - minimize player; unable to then control player

**Screenshot** - http://i.imgur.com/eP29652.png

MarshallOfSound:
I just tried your steps to reproduce.  And they didn't reproduce it :cry: 

Are you sure you aren't doing anything else in those steps?

t-readyroc:
OK, I've sort it out - you should be able to reproduce this:

1. open player
2. open side menu (3 bars, top-left)
3. minimize player

If you don't have the side menu open, the controls display correctly, but if the side menu is open when you minimize, they're f'ed.

MarshallOfSound:
It sounded like those steps would do it, but nope... I still can't reproduce this glitch.  Any other ideas?

t-readyroc:
OK, very odd.  It seems that if you do the above 3 steps _after_ you've already minimized once w/o menu successfully, it works.  **BUT** if you do the above 3 steps straight from a freshly opened player, you get the controls off-screen.  Definitely erratic.

MarshallOfSound:
I just did a clean run and did the following

1. Open the side nav
2. Play a song
3. Click the mini player button

And it works fine?  How consistently does it happen for you?

t-readyroc:
Every time.  But I suppose I have a workaround if you want to push this to the back of the queue.

I guess the only other thing to mention is that I go to My Library & hit the shuffle button to play, rather than picking a particular song.

jessecr:
I can reproduce this as well. Upon first opening the program I

1. Open the side nav
2. Click the mini player button

And the controls are off to the side (as seen in the screenshot in the initial comment) and I cannot return to the full view without quitting.

If I click the mini player button before opening the side nav, the controls will be visible as expected.

Windows 7 64-bit SP1

MarshallOfSound:
Goddamit, googles A/B testing....  If anyone can run the dev version and try figure out whats going on it would be much appreciated

jason-mehmel:
Also experiencing this on Debian Linux (Bunsen Labs).

[...]

PR #449, opened 2016-03-10
Fixed mini player controls being off-screen when the menu was expanded

N0ps32:
This should fix issue https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/issues/401

The problem was that the menu would push the controls off screen when it was expanded before one triggered the mini player mode. This PR closes the menu when entering mini mode therefore fixing the issue.


MarshallOfSound:
Thank so much for the fix, but can it be moved into a PR for the mini player library :+1: 

https://github.com/MarshallOfSound/gmusic-mini-player.js

Probably in this method  
https://github.com/MarshallOfSound/gmusic-mini-player.js/blob/master/src/gmusic-mini-player.js#L200

N0ps32:
I created a new PR in the other repository: https://github.com/MarshallOfSound/gmusic-mini-player.js/pull/5

Issue #403, opened 2016-03-09
Request - Add liked/disliked info to playback.json


spedrickson:
This would be very convenient for simple external programs like Rainmeter. Just a simple liked true/false and disliked true/false would be perfect.

Thanks!

MarshallOfSound:
For anyone wanting / willing to tackle this, it would be as simple as adding a couple of event listeners on the global `GPM` object inside `GPMInject` and firing events up to the PlaybackAPI.

An example is how the current implementation of `time` works

PR #451, opened 2016-03-11
Added rating event and rating information in JSON

N0ps32:
This PR fixes https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/issues/403

I added a new object called rating which has 2 fields liked and disliked that contain boolean values.

### DIFF
--- a/src/main/utils/PlaybackAPI.js
+++ b/src/main/utils/PlaybackAPI.js
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import createJSON from './_jsonCreator';
+import _ from 'lodash';
 
 class PlaybackAPI {
   constructor() {
@@ -24,6 +25,12 @@ class PlaybackAPI {
     Emitter.on('change:playback-time', (event, timeObj) => {
       this.setTime(timeObj.current, timeObj.total);
     });
+    // we throttle this function because of a bug in gmusic.js
+    // ratings are received multiple times here in a couple of ms
+    // to avoid writing the file 5+ times we throttle it to 500ms max
+    Emitter.on('change:rating', _.throttle((event, details) => {
+      this._setRating(details);
+    }, 500));
   }
 
   reset() {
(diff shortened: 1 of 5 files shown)


src/main/utils/PlaybackAPI.js
```diff
@@ -35,6 +42,10 @@ class PlaybackAPI {
         album: null,
         albumArt: null,
       },
+      rating: {
+        liked: false,
```
MarshallOfSound: Why are we translating a number in the 0-5 range into 2 booleans.  If users have the 5 star experiment turned on this could be any value (1, 2, 3, 4, 5)

src/main/utils/PlaybackAPI.js
```diff
@@ -63,12 +74,25 @@ class PlaybackAPI {
...
     this.data.songLyrics = null;
     this._fire('change:song', this.data.song);
     this._fire('change:lyrics', this.data.songLyrics);
     this._save();
   }
 
+  _setRating(rating) {
+    this.data.rating.liked = rating === '5';
+    this.data.rating.disliked = rating === '1';
+    this._fire('change:rating', this.data.rating);
```
MarshallOfSound: Implement this in the webSocket API as well please

MarshallOfSound:
Please add this to the docs for both the JSON API and the webSocket API

src/main/utils/PlaybackAPI.js
N0ps32: I didn't even know you could have more than like/dislike in GM.
I'll look into it.

N0ps32: I do not have this experiment maybe it has been removed?
According to this entry in the product forum it's gone for good: https://productforums.google.com/forum/#!topic/play/HwKmVF5N5YA

N0ps32:
I added the documentation and implemented the websocket API. I left the liked/disliked format for now since I don't think the 5 star system exists anymore.

N0ps32:
Is there a way to restart the appveyor build? It seems to have been stopped for some reason. The other ones pass though so the code should be ok.

MarshallOfSound:
Rebuilding now

PR #548, opened 2016-03-13
Window Titles Track Info

### LINKED ISSUE #387: Taskbar no longer shows Artist/track info on hover
OS: Windows 7 Enterprise (64-bit) SP1

Issue: Taskbar no longer shows Artist/track info on hover

Steps to Reproduce: Hover over the taskbar icon. The tooltip (including the previous/pause/next controls) displays, but rather than the artist/track info, just the "Google Play Music Desktop Player" text appears. I have the "Show song information always" setting checked. No hotkeys specified. Minimize to Tray is not set either.

gpmdp

jostrander:
Fixes #387 

Setting must be configured in the settings file, new property `enableWindowTitleInfo`.

### DIFF
--- a/src/main/features/core/windowTitleInfo.js
+++ b/src/main/features/core/windowTitleInfo.js
@@ -1,0 +1,17 @@
+const windowTitle = WindowManager.get(global.mainWindowID).getTitle();
+
+if (Settings.get('enableWindowTitleInfo')) {
+  PlaybackAPI.on('change:song', (songInfo) => {
+    WindowManager.get(global.mainWindowID).setTitle(
+      `${(songInfo.title || 'Unknown Song')} - ${(songInfo.artist || 'Unknown Artist')}`
+    );
+  });
+  PlaybackAPI.on('change:state', (state) => {
+    const songInfo = PlaybackAPI.currentSong();
+    WindowManager.get(global.mainWindowID).setTitle(
+      (state)
+      ? `${(songInfo.title || 'Unknown Song')} - ${(songInfo.artist || 'Unknown Artist')}`
+      : windowTitle
+    );
+  });
+}
--- a/src/main/configureBrowser.js
+++ b/src/main/configureBrowser.js
@@ -18,7 +18,7 @@ export default () => {
     autoHideMenuBar: true,
     frame: Settings.get('nativeFrame'),
     icon: path.resolve(`${__dirname}/../assets/img/main.png`),
-    title: 'Nucleus Player',
+    title: 'Google Play Music Desktop Player',
     nodeIntegration: true,
     'web-preferences': {
       preload: path.resolve(`${__dirname}/../inject/generic.js`),
--- a/src/main/features/core/index.js
+++ b/src/main/features/core/index.js
@@ -10,3 +10,4 @@ import './autoUpdater';
 import './websocketAPI';
 import './lyrics';
 import './applicationMenu';
+import './windowTitleInfo';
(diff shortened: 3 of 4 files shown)


MarshallOfSound:
> whoops, missed the main file...

This commit message is something I don't want to see in a PR, please squash your commits and force push

src/main/features/core/windowTitleInfo.js
```diff
@@ -0,0 +1,16 @@
+const windowTitle = WindowManager.get(global.mainWindowID).getTitle();
+let songInfo = {};
+
+if (Settings.get('enableWindowTitleInfo')) {
+  PlaybackAPI.on('change:song', (_songInfo) => {
+    songInfo = _songInfo;
```
MarshallOfSound: Why are we doing this?

src/main/features/core/windowTitleInfo.js
```diff
@@ -0,0 +1,16 @@
...
+    WindowManager.get(global.mainWindowID).setTitle(
+      (_songInfo.title && _songInfo.artist) ? `${songInfo.title} - ${songInfo.artist}` : windowTitle
```
MarshallOfSound: These two variables will **never** not be set

src/main/features/core/windowTitleInfo.js
jostrander: Saving the song info for state change events, such as play/pause so when the player is paused or stopped the player reverts to the regular title, and when playing again it changes to the Title - Artist. Could change to use PlaybackAPI calls directly.

src/main/features/core/windowTitleInfo.js
jostrander: Valid point, I needed to check if null, not if set. Will fix.

MarshallOfSound: They will never be null either?

src/main/features/core/windowTitleInfo.js
MarshallOfSound: Yep, I would prefer using `PlaybackAPI.getCurrentSong()` (or whatever it actually is)

src/main/features/core/windowTitleInfo.js
jostrander: They won't be null, but they can be empty. Just figured that one out. I have a few tracks without artist names, but if its that far out as a use case I'll just remove the check.

MarshallOfSound: Are those songs you uploaded your self?

I'd much rather go for
```js
(songInfo.title || 'Unknown Song')
(songInfo.artist || 'Unknown Artist')
```

kind of thing

jostrander:
Squashed it down to one commit, tried getting rid of the rebase commit but it wasn't happening. If that doesn't work feel free to clone this and commit it yourself however it's got to happen and I'll just delete my branch.

MarshallOfSound:
Have you tried running  
`git rebase -i origin/master`

You should then be able to mark all commits as "fixup"

jostrander:
I was trying to use the HEAD syntax on a soft reset, but I just wasn't going back far enough apparently, committing now

jostrander:
That didn't fix it either, even though my git log didn't show the rebase commit anymore. 

`git rebase -i origin/master` was of no help, probably would have been had I done it correctly in the first place.

jostrander:
Alright, commits squashed, does this all look good now?

MarshallOfSound:
Looks legit

PR #567, opened 2016-03-14
Fixed back button hiding with restored lastPage

MightyPork:
This PR fixes a bug introduced by remembering the last page and restoring to it on startup.
The Back button now goes to the home page ("Listen Now") if there's no history to use.

I also took the liberty to clean up the file and split it to functions, I hope you're okay with that. Better than having all the tweaks in a single block.

### DIFF
--- a/src/inject/GPMInject/interface/customUI.js
+++ b/src/inject/GPMInject/interface/customUI.js
@@ -1,6 +1,10 @@
 import { remote } from 'electron';
 import _ from 'lodash';
 
+
+// --- Helpers ---
+
+/** Hide elements by a selector */
 const hide = (elementSelector, kill = false) => {
   const nodeList = document.querySelectorAll(elementSelector);
   _.forEach(nodeList, (node) => {
@@ -12,6 +16,7 @@ const hide = (elementSelector, kill = false) => {
   });
 };
 
+/** Set CSS style by a selector */
 const style = (elementSelector, styleObject) => {
   const nodeList = document.querySelectorAll(elementSelector);
   _.forEach(nodeList, (node) => {
@@ -22,6 +27,7 @@ const style = (elementSelector, styleObject) => {
   });
 };
 
+/** Inject a CSS rule to the page (in a <style> tag) */
 const cssRule = (styles) => {
   const tag = document.createElement('style');
   tag.type = 'text/css';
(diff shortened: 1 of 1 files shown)


MightyPork:
uh I may have broken something. Wait a bit, will make sure this is right.

src/inject/GPMInject/interface/customUI.js
```diff
@@ -71,14 +90,29 @@ window.wait(() => {
...
+  const listenNowURL = 'https://play.google.com/music/listen#/now';
+
+  const backBtn = document.createElement('paper-icon-button');
+  backBtn.setAttribute('icon', 'arrow-back');
+  backBtn.setAttribute('id', 'backButton');
+  backBtn.setAttribute('class', 'x-scope paper-icon-button-0');
+  document.querySelector('#material-one-middle > sj-search-box').insertBefore(backBtn, null);
+
+  backBtn.addEventListener('click', () => {
+    const oldUrl = location.href;
```
MarshallOfSound: I've got a better idea for this code

```js
remote.getCurrentWindow().webContents.executeJavaScript('document.querySelector("webview").canGoBack()', false, (canGoBack) => {
    if (canGoBack) return history.back();
    location.href = listenNowUrl;
});
```

This is less vulnerable code (a settimeout relies on PC speed and network speed), this uses proper electron API's :+1:

MightyPork:
Update: Looks good. It was a bug on my home-use branch. Master is good.

src/inject/GPMInject/interface/customUI.js
MightyPork: I'll give that a go. Certainly more elegant :+1:

MightyPork: Works like a charm!

MarshallOfSound: Considering I wrote that on my phone while walking home I am quite impressed it just worked :+1:

MarshallOfSound:
Can we start getting into the habbit of squishing PR's into single commits.  It will make the history look a lot cleaner

MightyPork:
Right, done. I usually do that, but didn't think it's needed here. 

Ready for merge, if Travis doesn't make a scene.

Issue #232, opened 2016-03-06
Separate location on screen for mini player


FISHunderscore:
I noticed that in 3.0.0, the mini player now takes the place of the main window. This is annoying for users like myself who always have the mini player open on their secondary monitor, and want to be able to change what's playing without having to drag around windows.

If there is a way with Electron to make the mini player and main window store their locations separately, that would be awesome.

MarshallOfSound:
Should be able to do that

Maeldor:
+1

Can you also ensure it goes back to its previous size, please? I am having to resize it every time.

Thanks :)

FISHunderscore:
@jacobmischka I just meant store their locations separately. However, the ability to have two independent ones could be a good toggle setting.

MarshallOfSound:
@FISHunderscore two independent ones is a major task, two locations is much easier :+1:

Kun2210:
+1 
Having two locations would be more than good :+1:

MarshallOfSound:
Lol this was merged 5 days ago #closing

PR #585, opened 2016-03-14
Persist the size of the mini-player and store it independently of the main window.

SeanCline:
Switching between the mini-player and the regular window get tedious, especially when you prefer the mini player to be at the bottom of the screen. Issue #232 already tracks this.

This PR tracks the positions/sizes of the mini-player separately from the regular window and persists both across closes of the app.

P.S. Please review this carefully, I haven't done any dev against electron before.

### DIFF
--- a/src/inject/GPMInject/interface/mini.js
+++ b/src/inject/GPMInject/interface/mini.js
@@ -5,7 +5,6 @@ const webContents = mainWindow.webContents;
 const MINI_SIZE = 310;
 
 let mini = false;
-let oldSize;
 
 window.wait(() => {
   if (Settings.get('miniAlwaysShowSongInfo', false)) {
@@ -14,8 +13,13 @@ window.wait(() => {
 
   window.GPM.mini.on('enable', () => {
     Emitter.fire('mini', { state: true });
-    oldSize = remote.getCurrentWindow().getSize();
-    mainWindow.setSize(MINI_SIZE, MINI_SIZE);
+
+	// Restore the mini size/position from settings, otherwise use default size and regular position.
+    const miniSize = Settings.get('mini-size', [MINI_SIZE, MINI_SIZE]);
+    const miniPosition = Settings.get('mini-position', mainWindow.getPosition());
+    mainWindow.setSize(...miniSize);
+    mainWindow.setPosition(...miniPosition);
+
     mainWindow.setMaximumSize(MINI_SIZE, MINI_SIZE);
     webContents.executeJavaScript('document.body.setAttribute("mini", "mini")');
     remote.getCurrentWebContents().setZoomFactor(1);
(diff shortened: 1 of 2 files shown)


src/inject/GPMInject/interface/mini.js
```diff
@@ -14,8 +19,15 @@ window.wait(() => {
 
   window.GPM.mini.on('enable', () => {
     Emitter.fire('mini', { state: true });
-    oldSize = remote.getCurrentWindow().getSize();
-    mainWindow.setSize(MINI_SIZE, MINI_SIZE);
+
+	// Save regular size/pos and restore mini size/pos.
+    regularSize = remote.getCurrentWindow().getSize();
+    regularPosition = remote.getCurrentWindow().getPosition();
+    mainWindow.setSize(...miniSize);
```
MarshallOfSound: Because of the `persistAppState.js` logic, you will need to put  
```js
miniSize = Settings.get('mini-size', miniSize);
miniPosition = Settings.get('position', regularPosition);
```

src/inject/GPMInject/interface/mini.js
```diff
@@ -14,8 +19,15 @@ window.wait(() => {
...
+    if (miniPosition) {
```
MarshallOfSound: You can remove this if check, there is no case where miniPosition won't be defined (assuming you add the two recoomended lines above

src/inject/GPMInject/interface/mini.js
```diff
@@ -29,7 +41,13 @@ window.wait(() => {
     //      Same reason as specified in Electron src
     //        --> https://github.com/atom/electron/blob/master/atom/browser/native_window_views.cc
     mainWindow.setMaximumSize(99999999, 999999999);
-    mainWindow.setSize(...oldSize);
+
+    // Save mini size/pos and restore regular size/pos.
+    miniSize = remote.getCurrentWindow().getSize();
```
MarshallOfSound: You can get rid of these two lines if we are fetching the size from settings as per my above comment

SeanCline:
What do you think about taking it a step further and removing the globals `regularSize`, `regularPosition`, `miniSize`, and `miniPosition` in favour of always grabbing them from `Settings`?

MarshallOfSound:
I think that was what I was trying to get at, but the diff view on github made it quite hard to trace the usage of globals.  Technically it would be better to always grab from settings.  Test it and see if it works :+1:

MarshallOfSound:
OK, just took another look and it is looking good :+1: 

Can you just rebase and squash all your commits into a single commit and then I can merge it :+1:

SeanCline:
Alright. Squashed.

Right now, the only problem I see is that it's possible that we recover the window location somewhere that doesn't exist. For example, if the window (mini or full) is on another monitor, then the other monitor is removed....or if the resolution is lowed to the point that the window is off the monitor.

You can decide whether there needs to be some logic to constrain the window location to a valid display (probably checked through electron's `screen` api) before it is merged.

MarshallOfSound:
@SeanCline I have logic for handling that somewhere so I guess I'll extract that to a method and call it here as well.

:+1:

Issue #592, opened 2016-03-14
Javascript Error on Opening (Linux)


smilligan93:
**OS:**
Arch linux, aur package that pulls directly from the git repo [https://aur.archlinux.org/packages/google-play-music-desktop-player-git/](here)

**Issue Descriptions:**
Opening the program leads to
```
 Uncaught Exception:
TypeError: Cannot read property '0' of null
    at /usr/share/google-play-music-desktop-player/resources/app/build/index.js:97:19
    at Array.forEach (native)
    at EventEmitter.<anonymous> (/usr/share/google-play-music-desktop-player/resources/app/build/index.js:96:39)
    at emitOne (events.js:82:20)
    at EventEmitter.emit (events.js:169:7).
```

No windows appear when initially ran, just this popup. If I try to run the program again a window appears that is just black. Tried with '--disable-gpu' and still not working. The issue appears to be in the assignment of 'position' a couple lines above and is assigned null.

**Steps to Reproduce:**
Open the program on arch (unable to test on different distro).

**Side note**
I was able to get the program to open by putting 
```if (position != null) { 
_electron.screen.getAllDisplays().forEach(function (display) {
...
});
}````

Which then produces another js error on open:
```Uncaught Exception:
Error: Object has been destroyed
    at Error (native)
    at Object.<anonymous> (/usr/share/google-play-music-desktop-player/resources/app/build/main/features/core/keyboardShortcuts.js:17:1)
    at Module._compile (module.js:425:26)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:313:12)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/usr/share/google-play-music-desktop-player/resources/app/build/main/features/core/index.js:3:1)
    at Module._compile (module.js:425:26)```

but the app still opens and plays music, but unable to access "Desktop Settings" in the side bar.

MightyPork:
I can confirm this bug, it started with commit 1e241cb, before that it works OK.
All commits past that are broken for me.

![screenshot_2016-03-14_21-06-20](https://cloud.githubusercontent.com/assets/2041118/13758293/0825a640-ea29-11e5-81c5-7c65b54402c4.png)

If I could suggest something, why not make a "development" branch and put experimental stuff there, instead of straight to master?

[...]

PR #602, opened 2016-03-15
Workaround for #592, so it at least starts outside Windows

MightyPork:
Solves #592, so it starts on Linux without a nasty error dialog.


MarshallOfSound:
Seems fair enough,  it's a really strange electron bug.  I saw your comment, if I had to guess I'd put it down to bad GC?????

Oh well, if this fixes it :+1:

Issue #300, opened 2016-03-07
OSX: options to gray out and disable menubar icon


huyz:
The menubar icon for the player is the only one of mine that has color.  The OSX standard these days is grayscale.  Please provide a grayscale icon or an option to gray it out.
And also please provide an option to remove the menubar icon altogether.

MarshallOfSound:
The icon is now greyscale, PR to fix that was merged yesterday.

And as mentioned in other places the menubar is a core piece of functionality and removing it would make convuluted logic for the benefit of not very many users

huyl:
Hmmm, iTunes doesn't have a menubar icon.  How is that convoluted?

huyz:
Hmmm, iTunes doesn't have a menubar icon. How is that convoluted?

MarshallOfSound:
iTunes isn't this program, lots of functionality is very closely tied to the tray icon / menu.

huyz:
Cool, thanks for letting me know I should go back to Radiant Player

MarshallOfSound:
You're welome... :cactus: 

@chrismou You can have him

huyz:
Love the maturity haha

MarshallOfSound:
That maturity comment is beyond ironic....  I think I need to start adding community guidelines to this repository.

I think you don't understand @huyz this is not a business, I honstely don't care whether people use the program or not.  If you do **great** if you don't **woop woop**.  The app is here for the community, not individuals who want specific things and expect the maintainer/s to jump and perform....

huyz:
You're stating the obvious.  Open source wasn't invented yesterday.

Answering with "iTunes isn't this program" is just unnecessary.  I'll give you a few years to mature and figure it out.  Cheers, good luck.

MarshallOfSound:
You used a program as an example which is almost certainly written in a different language, on a different framework, with different goals...

This conversation is going nowhere, you seem to be under the impression my age affects my abilitiy / understanding of what you are saying....

Have fun with radiant...

Issue #609, opened 2016-03-15
On OS X, menu bar icon is low-res on retina screens.


nthapaliya:
Version 3.0.1 of this app, downloaded from the github releases page. I'm on El Capitan if that helps.

I searched through closed and open issues to see if this is a duplicate. I didn't find anything so I'm opening one now.

See attached screenshot.

<img width="139" alt="screen shot 2016-03-15 at 8 18 45 am" src="https://cloud.githubusercontent.com/assets/4131655/13780888/d45e4cc8-ea86-11e5-95cb-a79170d7738c.png">

jostrander:
@MarshallOfSound all that's needed here is an image that's 2x the normal size, and then the file name needs to be appended with `@2x`. If I had the sources for the graphics I'd make them but I can't upscale the current template logo. [see here](https://github.com/atom/electron/blob/master/docs/api/native-image.md#high-resolution-image)

jostrander:
I don't have mac to test these out, but if anyone wants to try them:

![mactemplate 5x](https://cloud.githubusercontent.com/assets/760989/13871101/1b5c6fbe-ecb8-11e5-9714-896be7b0bbed.png) `@5x`
![mactemplate 2x](https://cloud.githubusercontent.com/assets/760989/13871100/1b5bf3e0-ecb8-11e5-91c6-fcf92a2944b5.png) `@2x`
![mactemplate](https://cloud.githubusercontent.com/assets/760989/13871102/1b5cb9b0-ecb8-11e5-8bbb-efc426e7e4e8.png)

MarshallOfSound:
@jostrander I believe the current template images are just greyscale of the full size logo of which there is a 256x256 in the repo

amweiss-pki:
It looks like the sizes are a bit off so things are pixelated in some displays.
It should be the standard 16, 32, etc. sizes whereas they are currently quite odd sizes (19, 38, 95).

Do you have a vector image version somewhere? If so I can create the appropriate sizes.  Otherwise I can work from the 256 version to scale down to the correct sizes but they will still be pixelated.

MarshallOfSound:
No vector version's unfortunately, I might look into recreating it as a vector.

Best we have at the moment is the 256 one

PR #636, opened 2016-03-16
Add an option to disable the tray icon

amweiss:
I know #300 was closed out since you didn't want to submit to a single request, but I did the leg work for disabling the icon so hopefully this gets merged.  If you're open to having someone do the functionality but don't like this approach, let me know and I can modify the PR.

### DIFF
--- a/src/assets/less/dialog_window.less
+++ b/src/assets/less/dialog_window.less
@@ -39,6 +39,10 @@
       flex: 1;
       overflow: auto;
       padding: 0 40px;
+
+     .dependent {
+        margin-left: 40px;
+      }
     }
 
     .color-square-wrapper {
--- a/src/inject/settings/tray.js
+++ b/src/inject/settings/tray.js
@@ -10,3 +10,31 @@ if (Settings.get('minToTray', true)) {
 } else {
   $('#min-to-tray').removeAttr('checked');
 }
+
+$('#show-tray').change((e) => {
+  const checked = $(e.currentTarget).is(':checked');
+  Emitter.fire('settings:set', {
+    key: 'showTray',
+    value: checked,
+  });
+
+  $('#min-to-tray').attr('disabled', !checked);
+  if (!checked) {
+    Emitter.fire('settings:set', {
+      key: 'minToTray',
+      value: false,
+    });
+  }
+});
+
+if (Settings.get('showTray', true)) {
+  $('#show-tray').attr('checked', 'checked');
+  $('#min-to-tray').removeAttr('disabled');
+} else {
+  $('#show-tray').removeAttr('checked');
+  $('#min-to-tray').attr('disabled', 'disabled');
+  Emitter.fire('settings:set', {
+    key: 'minToTray',
+    value: false,
+  });
+}
--- a/src/main/features/core/tray.js
+++ b/src/main/features/core/tray.js
@@ -13,12 +13,6 @@ let audioDeviceMenu = [
   },
 ];
 
-if (process.platform === 'darwin') {
-  appIcon = new Tray(path.resolve(`${__dirname}/../../../assets/img/macTemplate.png`));
-} else {
-  appIcon = new Tray(path.resolve(`${__dirname}/../../../assets/img/main_tray.png`));
-}
-
 const setContextMenu = () => {
   const contextMenu = Menu.buildFromTemplate([
     { label: 'Show',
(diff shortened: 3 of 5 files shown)


src/main/features/core/applicationMenu.js
```diff
@@ -104,7 +104,7 @@ const template = [
 
 if (process.platform === 'darwin') {
   template.unshift({
-    label: 'Google Play Music Desktop Player',
+    label: (global.DEV_MODE)? 'Electron' : 'Google Play Music Desktop Player',
```
jostrander: Why are we switching to Electron for the label here when in dev mode?

amweiss-pki:
Looks like I missed some lint stuff, working on it now.

src/main/features/core/applicationMenu.js
amweiss-pki: On OSX is has to be the app name, and in dev mode it shows up as Electron.

src/main/features/core/applicationMenu.js
```diff
@@ -104,7 +104,9 @@ const template = [
 
 if (process.platform === 'darwin') {
   template.unshift({
-    label: (global.DEV_MODE)? 'Electron' : 'Google Play Music Desktop Player',
+    label: (global.DEV_MODE && process.platform === 'darwin') ?
```
MarshallOfSound: Couple of things here

1. The check for `darwin` is pointless because literally 2 lines above we are checking for `darwin`
2. `global.DEV_MODE` being `true` does **NOT** guarantee the application name is Electron or GPMDP.  We should use `app.getName()`
3. But most importantly this is overflowing the scope of this PR.  See the [CONTRIB.md](https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/master/CONTRIBUTING.md#pull-requests) :cry:

src/assets/less/dialog_window.less
```diff
@@ -38,6 +38,10 @@
       flex: 1;
       overflow: auto;
       padding: 0 40px;
+
+     * .dependent {
```
MarshallOfSound: This wildcard is not required and is detrimental to performance

src/inject/settings/tray.js
```diff
@@ -10,3 +10,20 @@ if (Settings.get('minToTray', true)) {
 } else {
   $('#min-to-tray').removeAttr('checked');
 }
+
+$('#show-tray').change((e) => {
```
MarshallOfSound: It looks like you haven't thought through the implications of this for all platforms.  For instance it appears at the moment on Windows that I can untick "Show in Tray" but still have "Minimize to Tray" ticked but disabled.

This means I now have NO way to quit the application.  It would just stay open always :-1:

src/main/features/core/tray.js
```diff
@@ -138,7 +142,9 @@ Emitter.on('audiooutput:list', (event, devices) => {
       });
     }
   });
-  setContextMenu();
+  if (appIcon) {
```
MarshallOfSound: Why do we have like 3 if statements in the same file...  Why can't all the code be inside one if statement

MarshallOfSound:
To be honest, I stand by my decision in #300 (and other issues), there is logic and process flows you haven't thought through in this PR.  One of which I mentioned (the unquittable windows application).  It is small gotchas like that, that make the tray icon required on lots of platforms.  

I'll leave it open to let you fix the issues with it and attempt to handle all the gotchas with applications quiting and then I'll look at it again.  But I'm not feeling it :cactus:

src/main/features/core/tray.js
amweiss: I was just preserving structure, but the one you highlighted is actually inside a function, not at the top level so it will have to remain.  I'll condense the others.

MarshallOfSound: Ah yeah, sorry, I just say a lot and commented on the one that was on screen :+1:

amweiss:
I can't get it building on windows right now, but pushing up some of the discussed fixes.

amweiss-pki:
Rebased on 3.0.1, if the opinion still stands close this out.  I can try to help make #609 better so at least the icon looks good if it's going to have to be there.

MarshallOfSound:
This is beyond out of date :laughing:  and I stand by my previous statements.  Closing this PR for now :+1:

PR #638, opened 2016-03-17
Ubuntu gnome media keys play/next/previous implementation.

### LINKED ISSUE #298: Media Keys in Linux -> MPRIS Implementation

jostrander:
This should fix the standard media keys on Ubuntu, uses DBus to capture the events coming from gnome without causing gnome to surrender the global listener. 

Partially fixes #298

### DIFF
--- a/src/main/features/linux/mediaKeysDBus.js
+++ b/src/main/features/linux/mediaKeysDBus.js
@@ -1,0 +1,24 @@
+import DBus from 'dbus';
+
+try {
+  const dbus = new DBus();
+  const session = dbus.getBus('session');
+
+  session.getInterface('org.gnome.SettingsDaemon', '/org/gnome/SettingsDaemon/MediaKeys',
+  'org.gnome.SettingsDaemon.MediaKeys', (err, iface) => {
+    if (!err) {
+      iface.on('MediaPlayerKeyPressed', (n, keyName) => {
+        switch (keyName) {
+          case 'Next': Emitter.sendToGooglePlayMusic('playback:nextTrack'); return;
+          case 'Previous': Emitter.sendToGooglePlayMusic('playback:previousTrack'); return;
+          case 'Play': Emitter.sendToGooglePlayMusic('playback:playPause'); return;
+          case 'Stop': Emitter.sendToGooglePlayMusic('playback:stop'); return;
+          default: return;
+        }
+        iface.GrabMediaPlayerKeys(0, 'org.gnome.SettingsDaemon.MediaKeys'); // eslint-disable-line
+      });
+    }
+  });
+} catch (e) {
+  // do nothing.
+}
--- a/circle.yml
+++ b/circle.yml
@@ -4,7 +4,7 @@ machine:
 
 dependencies:
   pre:
-    - sudo apt-get update; sudo apt-get install rpm fakeroot dpkg
+    - sudo apt-get update; sudo apt-get install rpm fakeroot dpkg libdbus-1-dev libglib2.0-dev
     - npm i -g npm
   cache_directories:
     - "~/.electron"
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "sinon-chai": "^2.8.0"
   },
   "optionalDependencies": {
+    "dbus": "^0.2.17",
     "electron-installer-debian": "^0.2.0",
     "electron-installer-redhat": "^0.2.0",
     "ll-keyboard-hook-win": "^2.0.1"
--- a/src/main/features/linux/index.js
+++ b/src/main/features/linux/index.js
@@ -1,0 +1,1 @@
+import './mediaKeysDBus.js';


MarshallOfSound:
Although a solid PR I wouldn't feel comfortable merging this without a LOT of try catches and error handling.  It seems like the kind of thing that wouldn't work on certain linux distros and would therefore through errors. 

I don't want those errors being sent to github.

src/main/features/linux/mediaKeysDBus.js
```diff
@@ -0,0 +1,17 @@
...
+const dbus = new DBus();
+const session = dbus.getBus('session');
+
+session.getInterface('org.gnome.SettingsDaemon', '/org/gnome/SettingsDaemon/MediaKeys',
+'org.gnome.SettingsDaemon.MediaKeys', (err, iface) => {
+  iface.on('MediaPlayerKeyPressed', (n, keyName) => {
+    switch (keyName) {
+      case 'Next': Emitter.sendToGooglePlayMusic('playback:nextTrack'); return;
+      case 'Previous': Emitter.sendToGooglePlayMusic('playback:previousTrack'); return;
+      case 'Play': Emitter.sendToGooglePlayMusic('playback:playPause'); return;
```
MarshallOfSound: Is there a `Stop` key as well?

jostrander:
Quit honestly I'm not sure what to catch, is there a way to catch all errors?

src/main/features/linux/mediaKeysDBus.js
jostrander: I don't have one so didn't think about it. I'll add it.

MarshallOfSound:
Doesn't appear to be an eventEmitter so a simple `trry { .. .} catch (e) { }` wrapping it should do.

Worth noting that these dependencies https://www.npmjs.com/package/dbus#dependencies

Should probably be added to the build process

jostrander:
Added try...catch, as well as build deps to circle.yml

MarshallOfSound:
Do they also need to be dependencies here  
https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/master/gulpfile.js#L222

jostrander:
I don't believe so, those dependencies both have -dev which leads me to believe they are only for building and don't need to be installed at run time.

jostrander:
Not to mention the mpris pull request I have also requries dbus and I've built and installed that without the dependencies.

src/main/features/linux/mediaKeysDBus.js
```diff
@@ -0,0 +1,24 @@
...
+          case 'Play': Emitter.sendToGooglePlayMusic('playback:playPause'); return;
+          case 'Stop': Emitter.sendToGooglePlayMusic('playback:stop'); return;
+          default: return;
+        }
+        iface.GrabMediaPlayerKeys(0, 'org.gnome.SettingsDaemon.MediaKeys'); // eslint-disable-line
+      });
+    }
+  });
+} catch (e) {
+  // do nothing.
```
MarshallOfSound: YOLO :+1:

MarshallOfSound:
LGTM now :+1:

ZenHarbinger:
I'm still not getting media key support on 16.04 64-bit.
Does mpris integration need to be done?
Is there a way to stream/log to the console to be sure that the module and/or keys are being registered?

I can see it compiling and linking w/ dbus.

jostrander:
Are you building from master? 

Try running `dbus-monitor interface=org.gnome.SettingsDaemon.MediaKeys` and pressing your media keys to see if you get a response.

Make sure you have the media keys set in Keyboard settings in Ubuntu otherwise it won't work.

ZenHarbinger:
I am building from master.
I am not getting any responses using the dbus-monitor.
I can however use the media keys in rhythmbox and I have written another audio app that does as well.
I do have the media keys set in the Keyboard settings.

ZenHarbinger:
Here is some python code I wrote that uses the media keys:
```
        self.properties = {
            ROOT_IFACE: self._get_root_iface_properties(),
            PLAYER_IFACE: self._get_player_iface_properties(),
        }
        bus_name = self._connect_to_dbus()
        super(MprisObject, self).__init__(bus_name, OBJECT_PATH)

        # init key handling
        logger.debug (u'Finding Spotipy...')
        session_bus = dbus.SessionBus();
        spotipy_bus = session_bus.get_object("org.mpris.MediaPlayer2.spotipy", "/org/mpris/MediaPlayer2")
        self.spotipy = dbus.Interface(session_bus, "org.mpris.MediaPlayer2.Player")

        logger.debug (u'Connect to MediaKeys...')
        bus = session_bus.get_object("org.gnome.SettingsDaemon", "/org/gnome/SettingsDaemon/MediaKeys")
        bus.connect_to_signal("MediaPlayerKeyPressed", self.key_pressed)
        self.mediakeys = dbus.Interface(bus, "org.gnome.SettingsDaemon.MediaKeys")
        self.mediakeys.GrabMediaPlayerKeys("spotipy", time.time())
```
https://github.com/ZenHarbinger/spotipy/blob/master/spotipy/frontends/mpris/objects.py

jostrander:
After doing some more research, the events only fire when a media player is attached. Try running rhythmbox or gpmdp and then running the dbus-monitor command. Alternatively if you want to try and debug yourself you can modify line 23 and do a `console.log(e);` and check the console for errors.

ZenHarbinger:
Works with rhythmbox, not with GPMDP, I'll work on the log checking in a bit, chores to do. :)

ZenHarbinger:
Previously, I've registered using MPRIS, does this need to be done?
So that DBUS knows the GDMDP is a media player?

jostrander:
Let me double check, I obviously have mpris set up, so let me rebase that branch and see if that fixes it.

jostrander:
Looks like GrabMediaPlayerKeys is failing to grab the keys, nothing to do with mpris. I'll troubleshoot further and see if I can come up with a solution.

jostrander:
Got it, I misplaced a closing bracket. I'll create another PR with the fix.

jostrander:
Fixed in #657

PR #640, opened 2016-03-17
About menu to show version from app.getVersion()

### LINKED ISSUE #578: "About" in Ubuntu
OS: Ubuntu 14.04.4
Version 3.0.1-1

There's no way---as far as I can see---to get information about the version of the player since there's no menu bar in Ubuntu as one might have on Windows or Mac OS X. You may think to add an "about" tab to the settings menu, or to the right-click menu on the tray icon. (Or a flag to the terminal command: "--version".)

jostrander:
Fixes #578

### DIFF
--- a/src/inject/core/appDetails.js
+++ b/src/inject/core/appDetails.js
@@ -4,7 +4,6 @@ import path from 'path';
 
 window.addEventListener('load', () => {
   if (window.$ && Settings.get('welcomed') !== remote.app.getVersion()) {
-    $('[data-app-version]').text(remote.app.getVersion());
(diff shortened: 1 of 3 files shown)


src/inject/core/appDetails.js
```diff
@@ -14,4 +14,14 @@ window.addEventListener('load', () => {
       },
     });
   }
+  if (window.$) {
+    const modal = $('#about');
+    $('[data-app-version]').text(remote.app.getVersion());
```
MarshallOfSound: If we are adding this line can we delete line 7.  It would result in duplicate processing

MarshallOfSound:
One minor thing, otherwise looks good

jostrander:
Updated

PR #657, opened 2016-03-17
Fix bug in media player keys ubuntu

jostrander:
I don't know when this happened, but this fixes the bug of them not working in the master branch.

### DIFF
--- a/src/main/features/linux/mediaKeysDBus.js
+++ b/src/main/features/linux/mediaKeysDBus.js
@@ -15,8 +15,8 @@ try {
           case 'Stop': Emitter.sendToGooglePlayMusic('playback:stop'); return;
           default: return;
         }
-        iface.GrabMediaPlayerKeys(0, 'org.gnome.SettingsDaemon.MediaKeys'); // eslint-disable-line
       });
+      iface.GrabMediaPlayerKeys(0, 'org.gnome.SettingsDaemon.MediaKeys'); // eslint-disable-line
     }
   });
 } catch (e) {


MarshallOfSound:
:cactus: :clap: 

I'll take your word for it #trust

ZenHarbinger:
OK, I figured it out; it has to do with the version of node.

Does NOT work w/ node 4.4.0 (previous testing was with 4.3.4), but DOES with 5.9.0.
I just did a new build with a re-downloaded master and this is what seems to be the case.
When running w/ 4.4.0, I do get a warning about: 
`[Error: Module version mismatch. Expected 47, got 46.]`
After the Last.fm api key output warning.
No such warning exists with node 5.9.0
I am not sure if these are related, but I can say that dbus/play/pause/prev/next do work when using 5.9.0.

MarshallOfSound:
Hmmmm, that is really strange.  Electron is shipping with node `5.X` so it shouldn't be an issue in the packaged app.  But the fact that the LTS version of node doesn't support something is strange.

Gonna look into that.

jostrander:
Just to confirm, can you do a fresh clone into a new directory, then npm install, build, and start via 4.4.0?

ZenHarbinger:
Sure, but now I have a question.
Node.js 4.4 is recommended for dev,
But when downloading node.js 4.4, it comes w/ npm 2.14.20
Do I need to mix these?

MarshallOfSound:
You need to run `npm i -g npm` to get the latest version of NPM

I reccomend node `4.4.X` for LTS reasons and NPM `3.x` because it flattens dependencies much better and this fixes several issues on Windows.

ZenHarbinger:
OK, I completely removed node 5.9.0
I re-installed a fresh copy of 4.4.0

Here is my log:
I did a clean grab of the source.
I did an update to npm
then install/run build/start
No media buttons is the end result for me.

[build.log.txt](https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/files/179825/build.log.txt)

jostrander:
Can confirm, did exact steps and got exact result. Will debug further since this is the recommended node version presently.

MarshallOfSound:
From a cursory search it appears as though the actual act of changing node versions is causing this issue.  You need to rebuild dependencies when you change node versions, the easiest way to do this is to delete `node_modules` and run `npm i` again.

ZenHarbinger:
I didn't even have a node_modules directory when I ran.  As I had just done a clean clone of the source.

MarshallOfSound:
Do you have any global node modules that are staying around?

jostrander:
Running install from node v4.4.0 without a node_modules folder, I still
received the error. I even tried npm rebuild.

ZenHarbinger:
I removed my ~/.npm and ~/.node-gyp directories for what it's worth
clean grab
same result (no media keys).

MarshallOfSound:
Hmmm, does anyone know which module is causing the error? (My ubuntu VM isn't starting atm so I can't really do any testing :cry: )

jostrander:
Looks like it's node-dbus. If I remove the files that import that module it stops reporting the error.

jostrander:
I'm not sure why, but it just seems that `dbus.node` is not being compiled for the right module version. The offending import line is `node_modules/dbus/lib/dbus.js:3:13` and that line is:
```
var _dbus = require('../build/Release/dbus.node');
```
which is this file:

```
node_modules/dbus/build/Release/dbus.node: ELF 64-bit LSB  shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=bf7efe8d40352aea8cf05f94ea95575d9441804a, not stripped
```

Original stack trace:
```
Error: Module version mismatch. Expected 47, got 46.
    at Error (native)
    at process.module.(anonymous function) [as dlopen] (ATOM_SHELL_ASAR.js:157:20)
    at Object.Module._extensions..node (module.js:450:18)
    at Object.module.(anonymous function) [as .node] (ATOM_SHELL_ASAR.js:157:20)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:313:12)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (~/Google-Play-Music-Desktop-Player-UNOFFICIAL-/node_modules/dbus/lib/dbus.js:3:13)
    at Module._compile (module.js:425:26)
```

jostrander:
@ZenHarbinger I think I figured it out, but can you please confirm? I used node v4.4.0, and npm v3.8.2 via nvm, removed all ubuntu apt packages, however probably irrelevant. 

What I did was `npm install && npm run build && npm start` and confirmed the error was present.

I then ran `./node_modules/.bin/electron-rebuild` and it gave no output, however running `npm start` didn't give any module error output and media keys were working. Let me know if this fixes it for you and I'll do a pull request that should run `electron-rebuild` on `npm install` or similar.

ZenHarbinger:
Yes, everything was as you expected.
`./node_modules/.bin/electron-rebuild` ran w/o output and then `npm start` ran w/o errors and media keys working.

Using npm 3.8.2 and node 4.4.0.

MarshallOfSound:
@jostrander I'll set up the electron-rebuild scripts for linux  but for future reference here is the code I have to rebuild on windows

https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/tree/master/vendor

and here is the information about using native modules with electron

https://github.com/atom/electron/blob/master/docs/tutorial/using-native-node-modules.md

jostrander:
@MarshallOfSound Sounds good :+1: I took a peek in /vendor/ which is what gave me a hint to try the electron-rebuild command.

MarshallOfSound:
Yeah, I would have clicked onto the rebuild but I didn't catch this was a native module :cry:

jostrander:
Did you fix this in master yet @MarshallOfSound, looks like I still have to run electron-rebuild.

MarshallOfSound:
Not fixed yet, spent some time on theming yesterday.

There is some interesting things I need to do in the gulp file to make this work :/

jostrander:
Can't do it as an npm script such as this?

```
"scripts": {
    "install": "electron-rebuild"
}
```

MarshallOfSound:
Not quite as simple as that unfortunately.  For instance, that command does not work on OS X

Also, we have to be able to build 32bit and 64but packages during the build process.  This requires args being passed to electron-rebuild

jostrander:
Fair enough!

rijnhard:
@MarshallOfSound apologies for the necro, but was this fixed? Just installed the latest GPMDP on Linux Mint 17.3 (Ubuntu 16.04 underneath) and Media keys aren't working, `dbus-monitor` shows nothing. Oddly enough it does appear in the Cinnamon taskbar though.

jostrander:
Install method? Those are two different portions of code. One is mpris the other is dbus media keys. That code really hasn't been touched in a while and I wrote it on Ubuntu 16.04, so not sure what would be at fault here.

rijnhard:
@jostrander x64 .deb from official site.

Issue #649, opened 2016-03-17
We should have a log file for uncaught exceptions


MarshallOfSound:
We should store a file somewhere which maintains the logs of the application.  This should include but should not be limited to

* [x] Any errors that occur (uncaught exceptions)
* [x] Transitioning between phases of  the update process and any errors that occur during that.
* [x] Update server connection failed
* [x] Start up and shut down events
* [x] Keybinding failures

PR #740, opened 2016-03-23
Bring in the `winston` package for logging.

SeanCline:
In case you are looking for contributions for issue #649, here is a preliminary take at logging support. I'm not sure if this is what you were thinking in terms of logging. If not, I'll close the PR.

(Mostly, I just wanted to mess around with winston for logging since I'd never used it.)

### DIFF
--- a/src/inject/GPMInject/index.js
+++ b/src/inject/GPMInject/index.js
@@ -35,7 +35,7 @@ const waitForExternal = setInterval(() => {
       try {
         fn();
       } catch (e) {
-        console.error(e); // eslint-disable-line
+        Logger.error('Emitter fn() threw exception.', e);
       }
     });
 
--- a/src/inject/GPMInject/index.js
+++ b/src/inject/GPMInject/index.js
@@ -3,6 +3,10 @@ import { remote } from 'electron';
 
 import '../generic';
 
+// Initialize the global Logger to forward to the main process.
+window.Logger = remote.getGlobal('Logger');
+Logger.debug('Renderer process logger initialized.');
+
 // DEV: Hold all Emitter events until the GPM external assets have been loaded.
 Emitter.ready = false;
 const waitingQueue = [];
--- a/src/inject/GPMInject/interface/voiceControls/SpeechRecognizer.js
+++ b/src/inject/GPMInject/interface/voiceControls/SpeechRecognizer.js
@@ -66,7 +66,7 @@ export default class SpeechRecognizer {
     if (matchedFn) {
       matchedFn(fnArg)
         .catch((error) => {
-          console.info(error);
+          Logger.info(error);
         })
         .then((unusedCommand) => {
           if (unusedCommand === undefined) {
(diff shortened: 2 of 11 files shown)


MarshallOfSound:
I only took a quick glance, need to sleep soon.  Couple of comments but overall pretty much EXACTLY :+1:  what I was planning on doing :smile:

src/inject/GPMInject/index.js
```diff
@@ -35,7 +35,7 @@ const waitForExternal = setInterval(() => {
       try {
         fn();
       } catch (e) {
-        console.error(e); // eslint-disable-line
+        Logger.error('Emitter fn() threw exception.', e);
```
MarshallOfSound: Anything inside "inject" won't have the global logger availible, electron works in two processes (renderer and main), look into this to figure out how you would make them global in both.

There is also a trick in electron that would go something like `require('electron').remote.getGlobal('globalNameHere')`

src/inject/GPMInject/interface/voiceControls/SpeechRecognizer.js
```diff
@@ -95,7 +95,7 @@ export default class SpeechRecognizer {
 
     _.forEach(actions, (actionString) => {
       if (this._handlers[actionString.toLowerCase()]) {
-        console.error(`The action "${action.toLowerCase()}" already has a handler registered`); // eslint-disable-line
+        Logger.error(`The action "${action.toLowerCase()}" already has a handler registered`); // eslint-disable-line
```
MarshallOfSound: I would want this to go to the console aswell, not sure if that is how winston works or not

src/main/constants.js
```diff
@@ -3,9 +3,5 @@ export const LASTFM_API_SECRET = process.env.GPMDP_LASTFM_API_SECRET;
 
 if (!LASTFM_API_KEY || !LASTFM_API_SECRET ||
   LASTFM_API_KEY === 'undefined' || LASTFM_API_SECRET === 'undefined') {
-  console.error('####################################'); // eslint-disable-line
-  console.error('############### ERROR ##############'); // eslint-disable-line
-  console.error('######### LASTFM  API KEYS #########'); // eslint-disable-line
-  console.error('######## HAVE NOT BEEN SET #########'); // eslint-disable-line
-  console.error('#####################################'); // eslint-disable-line
+  Logger.error('Last.fm API keys have not been set!');
```
MarshallOfSound: I would want this to go to the console aswell, not sure if that is how winston works or not

src/main/features/core/autoUpdater.js
```diff
@@ -57,7 +64,7 @@ const checkUpdateServer = () => {
   https.get('https://update.gpmdp.xyz', () => {
     setUpAutoUpdate();
   }).on('error', () => {
-    console.log('################### !! Update server down !! ##################'); // eslint-disable-line
+    Logger.error('################### !! Update server down !! ##################');
```
MarshallOfSound: I would want this to go to the console aswell, not sure if that is how winston works or not

SeanCline:
I'll address the issue with the global logger not being available in the renderer process later today.

Winston lets you select how verbose you want each of the transports to be. I chose to make it so only errors go to console and anything info level or higher goes to log. And, in dev mode, debug or higher go to console. This is very much open to tweaking.

PR #855, opened 2016-03-30
Add AudioVideo Audio to .desktop builds in linux

### LINKED ISSUE #853: Linux build uses wrong category in the .desktop file
This is just a proposal to change the application category from Utilities to Multimedia (which I think makes more sense) in the .desktop file for the Linux build.

jostrander:
Fixes #853


jostrander:
@MarshallOfSound this follows the freedesktop.org .desktop file spec:

https://standards.freedesktop.org/menu-spec/latest/apa.html

PR #872, opened 2016-04-01
No hyphen

jostrander:
There's no hyphen in disliked. Sorry to nitpick.


MarshallOfSound:
But  
But  
Autocorrect plz....

PR #884, opened 2016-04-03
Add button in now playing menu for lyrics.

jostrander:
Also modified the lyrics screen to fade in at 150ms, seemed a little less harsh after clicking the menu button.

### DIFF
--- a/src/inject/GPMInject/interface/customUI.js
+++ b/src/inject/GPMInject/interface/customUI.js
@@ -99,6 +99,51 @@ function installDesktopSettingsButton() {
   document.querySelectorAll('.nav-section.material')[0].insertBefore(dSettings, document.querySelectorAll('.nav-section.material > a')[2]); // eslint-disable-line
 }
 
+/* eslint-disable max-len, no-multi-str */
+function installNowPlayingButton(text, id, callback) {
+  const ripple = '<paper-ripple style="-webkit-user-select: none;"> \
+                  <div id="background" class="style-scope paper-ripple" style="-webkit-user-select: none;"></div> \
+                  <div id="waves" class="style-scope paper-ripple" style="-webkit-user-select: none;"></div> \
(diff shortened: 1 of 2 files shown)


src/inject/GPMInject/interface/customUI.js
```diff
@@ -99,6 +99,51 @@ function installDesktopSettingsButton() {
   document.querySelectorAll('.nav-section.material')[0].insertBefore(dSettings, document.querySelectorAll('.nav-section.material > a')[2]); // eslint-disable-line
 }
 
+/* eslint-disable max-len, no-multi-str */
+function installNowPlayingButton(text, id, callback) {
+  const ripple = '<paper-ripple style="-webkit-user-select: none;"> \
```
MarshallOfSound: If you use backticks to surround the string you can go multi line without "\"

src/inject/GPMInject/interface/customUI.js
```diff
@@ -99,6 +99,51 @@ function installDesktopSettingsButton() {
...
+                  <div id="background" class="style-scope paper-ripple" style="-webkit-user-select: none;"></div> \
+                  <div id="waves" class="style-scope paper-ripple" style="-webkit-user-select: none;"></div> \
+                  </paper-ripple>';
+  const content = document.createElement('div');
+  content.setAttribute('class', 'goog-menuitem-content');
+  content.innerHTML = `${ripple} ${text}`;
```
MarshallOfSound: With lot's of stuff you don't need to add in the ripple HTML, GPM somehow makes it for you.  Examples are the buttons we inject.  Did you try without adding the ripple?

jostrander: Just removed the ripple html, and no ripple :-1:

MarshallOfSound: :cry:

PR #887, opened 2016-04-03
Fix Lyrics XSS

### LINKED ISSUE #886: Lyrics API can perform XSS attack.
[The recent added lyrics code](https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/commit/ea917df658e4ce447cae870c6ce4dea40ae988ff#diff-1f78092df3330f593ead79d0d745d4deR18) opens the user to an XSS attack by the lyrics API. There is no reason to have this risk. Instead the received lyrics should be escaped properly.

@MarshallOfSound

jostrander:
Fixes #886, imports an XSS library that does a pretty good job of removing everything except whitelisted tags, including parameters such as onclick, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "semver": "^5.1.0",
     "winston": "^2.2.0",
     "ws": "^1.0.1",
+    "xss": "^0.2.10",
     "yargs": "^4.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


kevincox:
This looks good to me!

PR #909, opened 2016-04-06
Add function to pause after current song

### LINKED ISSUE #445: Feature request: Stop playing after current song
This is something I really miss from practically all players that don't have it, and I would love to see it here, and I'm more hopeful that you'll listen to my wishes than Google. :)

Maybe through a menu by holding the pause-button? Or just somewhere else, doesn't really matter. There are just so many times that you want to finish listening to the current track but then stop after that. It can be circumvented by deleting everything afterwards in the current queue, but a dedicated function would be so much better.

MCManuelLP:
Sometimes it's off by a second, I can't seem to fix that...


jostrander:
Well, you beat me to it! I have this half done in a branch locally, guess I'll have to clone yours and try it out... :+1:

src/inject/GPMInject/interface/toast.js
```diff
@@ -0,0 +1,26 @@
...
+  disableButton.addEventListener('click', (e) => {
+    remote.getGlobal('PlaybackAPI')._setPauseAfter(false);
+    e.preventDefault();
+    return false;
+  });
+  toast.appendChild(disableButton);
+  document.body.appendChild(toast);
+  toast.show();
+  Emitter.on('toastHide:pauseAfter', () => {
+    toast.hide()
```
jostrander: We're hiding the toast here, but the toast is being created via new DOM element, so the old one is never destroyed. Currently I have 3 paper-toasts. Can we optimize this to reuse the same element?

jostrander:
While you're there, can you add `bottom: 102px` to the style attribute? It currently is covering up the track info, and that will put it above that.

jostrander:
@MarshallOfSound can you look over this and see if you notice anything that might cause #910?

src/main/utils/PlaybackAPI.js
```diff
@@ -134,6 +138,17 @@ class PlaybackAPI {
     this._save();
   }
 
+  _setPauseAfter(mode) {
+    if (mode) {
+      Emitter.sendToGooglePlayMusic('toast:pauseAfter', null);
```
jostrander: can we change the scope here from `toast:` to `pauseAfter:show` and `pauseAfter:hide`? Seems like it might make more sense in the long run.

src/inject/GPMInject/interface/toast.js
```diff
@@ -9,6 +9,7 @@ Emitter.on('toast:pauseAfter', () => {
   disableButton.innerHTML = 'Don\'t pause after this song';
   disableButton.style.color = '#E53935';
   disableButton.style.margin = '0';
+  disableButton.style.marginBottom = '102px';
```
jostrander: `bottom` not `marginBottom`, as it's absolutely positioned in the lower left corner.

src/main/utils/PlaybackAPI.js
```diff
@@ -134,6 +138,17 @@ class PlaybackAPI {
     this._save();
   }
 
+  _setPauseAfter(mode) {
```
MarshallOfSound: This method should not be in `PlaybackAPI`.  The playback API exposes information about the player to an external controller, through either the JSON api or the WebSocket API.  This appears to be a method that controls the player so should not really be here.

src/main/utils/PlaybackAPI.js
```diff
@@ -134,6 +138,17 @@ class PlaybackAPI {
...
+    if (mode) {
```
MarshallOfSound: The parameter "mode" is very non descript.  I have no idea what it is by looking at it.  You may have copied from repeat or shuffle functions but for those a "repeat mode" and a "shuffle mode" is actually a thing.

src/main/utils/PlaybackAPI.js
```diff
@@ -134,6 +138,17 @@ class PlaybackAPI {
...
+      Emitter.sendToGooglePlayMusic('pauseAfter:show', null);
+    } else {
+      Emitter.sendToGooglePlayMusic('pauseAfter:hide', null);
+    }
+    this.data.pauseAfter = mode;
```
MarshallOfSound: Again, this is not the kind of thing we would want inside the PlaybackAPI

src/main/utils/PlaybackAPI.js
```diff
@@ -42,6 +42,9 @@ class PlaybackAPI {
     Emitter.on('change:repeat', _.throttle((event, mode) => {
       this._setRepeat(mode);
     }), 20);
+    Emitter.on('change:pauseAfter', _.throttle((event, mode) => {
```
MarshallOfSound: There is no need to throttle this event, it will only be fired one at a time

src/inject/core/pauseAfter.js
```diff
@@ -0,0 +1,21 @@
...
+// to avoid play pause to be called 5+ times we throttle it to 500ms max
+window.addEventListener('load', _.throttle(() => {
+  remote.getGlobal('PlaybackAPI').on('change:song', () => {
+    if (remote.getGlobal('PlaybackAPI').currentPauseAfter()) {
+      try {
+        GPM.playback.playPause();
+        remote.getGlobal('PlaybackAPI')._setPauseAfter(false);
+      } catch (err) {
+// calling GPM throws a ReferenceError for whatever reason, but it works, so...
+        if (err.name !== 'ReferenceError') {
```
MarshallOfSound: GPM throws a reference error because you put this in the wrong folder.  If you put it in the `GPMInject` folder it will only be injected into the Google Play Music webview not every single rendering process

src/inject/GPMInject/interface/toast.js
```diff
@@ -0,0 +1,27 @@
+import { remote } from 'electron';
+
+Emitter.on('pauseAfter:show', () => {
+  const toast = document.createElement('paper-toast');
```
MarshallOfSound: From memory we have another bit of code somewhere that also makes a toast (I think in the error handler) perhaps as part of this PR (to reduce code duplication) you could extract the creation of a toast to a method somewhere.

MCManuelLP: Hmm, it would have to be a method with loads of parameters... There are very few things absolutely the same...

MCManuelLP: Also, general question, is the toast link colored for you? It doesn't seem to be for me, but maybe that's just me...

src/inject/GPMInject/interface/pauseAfter.js
```diff
@@ -0,0 +1,31 @@
+import { remote } from 'electron';
+
+let pauseAfter = false;
+
+window.addEventListener('load', () => {
+  remote.getGlobal('PlaybackAPI').on('change:song', () => {
```
MarshallOfSound: Inside GPM inject it makes more sense to use `GPM.on('change:son')`

MCManuelLP:
Well, finally got around doing all changes, sorry, took a while...

MarshallOfSound:
I like it with the changes :+1: 

Thanks for the contrib

Issue #919, opened 2016-04-08
Cinnamon sound applet low res album art and missing progress bar


wengole:
Thanks for your help so far.

Two (relatively) minor things wrong with what the cinnamon applet shows is a lack of track progress bar, and the album art is low res.

The former I believe is another case of incorrect typing of the track length, as #906 had snippets of dbus data sent from Banshee and GPMDP and there was a definite discrepancy in the type.

I'm not sure the latter is even fixable, unless Google provides multiple image sizes.

jostrander:
The progress bar showing up is an issue with the applet itself. If you don't mind, take a look into the JS for the applet.
The one I'm familiar with is here:
https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/sound%40cinnamon.org/applet.js#L25
but yours may be different.
Currently it looks like there is a whitelist for players that support the progress bar, and its hard coded. Can you check this out on your system? The path might be similar.

However the progress bar once whitelisted doesn't work, so it probably does need a datatype adjustment. I'll look into that.

The latter can probably be fixed. What size would you need? I can probably safely provide around 1024x1024px, but I'd prefer to keep in in the middle around 512x512 to keep the size low. Currently I believe its 100px;

wengole:
I'll look into the progress bar issue.

Image size, I think 512px would indeed be best, at least double what it is now though.

jostrander:
Added 300px album art, 512 seemed a bit excessive.

jostrander:
As far as the track length, unfortunately it looks like there's no way to marshall into int64 from Javascript as Javascript doesn't support 64 bit integers natively.

MarshallOfSound:
@jostrander That 64bit integer thing should probably be handled by a c++ native module (node-dbus) probably.

jostrander:
v8 doesn't support checking for 64 bit integers it appears, so even a native module won't work in this case, but maybe I'm wrong as I don't play with C++ too much.

http://bespin.cz/~ondras/html/classv8_1_1Value.html

MarshallOfSound:
I'm pretty sure that `Integer::Value()` returns a 64bit integer in V8

jostrander:
[...]

PR #928, opened 2016-04-09
MPRIS Album Art Size

jostrander:
Partial Fix to #919.


jostrander:
I have no idea how I managed to repeat that commit into this PR...

PR #929, opened 2016-04-10
Initial gulp re-factoring for separate Linux builds.

jostrander:
Not even going to lie, I'm cheating and using this PR so CI will validate for all three build environments at once :+1:

Do not merge.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   "homepage": "http://www.googleplaymusicdesktopplayer.com",
   "license": "MIT",
   "dependencies": {
-    "electron-prebuilt": "0.37.5",
+    "electron-prebuilt": "0.37.6",
     "html-entities": "^1.2.0",
     "lastfm": "^0.9.2",
     "lodash": "^4.0.0",
--- a/src/inject/core/index.js
+++ b/src/inject/core/index.js
@@ -1,7 +1,11 @@
-import './control-bar';
 import './goToURL';
 import './uninstall';
-import './webviewLoader';
 import './updateAvailable';
 import './appDetails';
 import './lyrics';
+
+// Anything that requires the DOM needs to go here as of Electron 0.36.6
+document.addEventListener('DOMContentLoaded', () => {
+  require('./webviewLoader');
+  require('./control-bar');
+});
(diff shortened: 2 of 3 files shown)


jostrander:
@BrianAllred what works best for an AUR, the files straight out of the packager, or a deb/rpm? For example:

image

MarshallOfSound:
@jostrander Just going to question the use of `asar`.  It has quite a few caveats that I believe would break our current usage of the `fs` module.

http://electron.atom.io/docs/v0.36.0/tutorial/application-packaging/#treating-an-asar-archive-as-a-normal-file

jostrander:
I did read up on that, and I'm not having any immediate issues with an `asar` version of the build. Do you have any examples of things that would fail with the `fs` module? The example you linked to is showing that `original-fs` is needed only to read the `asar` file instead of its contents. However it did cause windows to fail to build because it couldn't grab something with the keyboard hooks.

jostrander:
Also I believe there are some npm modules that are being packed into the resources folder that don't need to be, but the only way I've seen that fixed is by having two package.json files. One for development, and one for building.

MarshallOfSound:
@jostrander We have `prune` set to `true` so no development npm modules are being included in the package

jostrander:
Yep what I was seeing was the optional deps Debian and redhat packager npm packages in the resources directory. If that's the only few that make it in I'm not so concerned about it.

BrianAllred:
@jostrander The files are much preferred. That keeps the user from having to install dpkg and rpmtools as build dependencies.

tf2manu994:
@jostrander Just letting you know that circleci didn't like that last commit

jostrander:
Yeah I saw, haven't had time to fix it. I might have to implement some kind of flag system to figure out what needs to build and when, because currently make:linux rebuilds 4 times due to dependencies for each 32 and 64 and deb and rpm. Will look into it soon.

MarshallOfSound:
If you use `gulp` "requires" syntax correctly it should only build linux 32bit once and 64bit once.  Gulp has magical ways of determining task dependencies.  If you set it up right it handles it all for you @jostrander

jostrander:
Yeah I know, it did figure out quite a bit, but there are still 4 builds unfortunately, compared to the 16 I calculated.  The last thing I tried to commit was zipping up the unpacked folders, but zip was throwing an error so I'll have to test that some more.

PR #946, opened 2016-04-15
Separate Linux builds/packages and some refactoring.

### LINKED ISSUE #894: Allow for specific linux builds.
The build tooling should allow a package to be built for any one specific distribution so specific packages can be built without building for both architectures. Thoughts? This would specifically help the Arch community out so they can clone, build and update via AUR without needing both architecture libraries available.

jostrander:
Accidentally closed other PR and couldn't reopen :door:.

Fixes #894


BrianAllred:
So the proper syntax will be 

```
npm run package:linux:32 
```
for 32-bit and
```
npm run package:linux:64
``` 
for 64-bit?

gulpfile.babel.js
```diff
@@ -292,83 +363,115 @@ gulp.task('rpm:linux', ['package:linux'], (done) => {
...
+    src: `dist/${packageJSON.productName}-linux-x64`,
+    arch: 'amd64',
+  }), (err) => {
+    console.log('32bit rpm package built');
+    if (err) return done(err);
+    done();
+  });
+});
+
 gulp.task('make:linux', ['deb:linux', 'rpm:linux'], (done) => {
```
MarshallOfSound: Hmmmm, this line might not work in my opinion.

Gulp will fire off both the `deb:linux` and `rpm:linux` tasks simultaneously.  Both those tasks run 32bit/64bit rebuilds of native modules.  Depending on how quick each task runs a 32bit build might have been rebuilt into 64bit before the copy process.  We should use something like this https://www.npmjs.com/package/run-sequence (until gulp 4.0 is released) that will force execution order.

jostrander: Err, that's existing code used to build via circle CI, unmodified. Did you mean another line? I didn't  do anything that would mess with the normal CI process this time. Only added the 6 Linux only build commands.

MarshallOfSound: Yeah I know it is unmodified, it was more of a general comment.  It was probably an issue before this PR but reviewing code catches things that writing it doesn't :+1: :laughing:

gulpfile.babel.js
```diff
@@ -213,44 +234,69 @@ gulp.task('make:darwin', ['package:darwin'], (done) => {
...
+const debianDefaults = {
+  bin: packageJSON.productName,
+  dest: 'dist/installers/debian',
+  depends: ['libappindicator1'],
+  maintainer: 'Samuel Attard <samuel.r.attard@gmail.com>',
+  homepage: 'http://www.googleplaymusicdesktopplayer.com',
+  icon: 'build/assets/img/main.png',
+  categories: ['AudioVideo', 'Audio'],
+};
 gulp.task('deb:linux', ['package:linux'], (done) => {
```
MarshallOfSound: This should sequentially run `deb:linux:32` and `deb:linux:64`

gulpfile.babel.js
```diff
@@ -261,27 +307,52 @@ gulp.task('deb:linux', ['package:linux'], (done) => {
...
+  bin: packageJSON.productName,
+  dest: 'dist/installers/fedora',
+  depends: ['libappindicator1'],
+  maintainer: 'Samuel Attard <samuel.r.attard@gmail.com>',
+  homepage: 'http://www.googleplaymusicdesktopplayer.com',
+  icon: 'build/assets/img/main.png',
+  categories: ['AudioVideo', 'Audio'],
+};
+
 gulp.task('rpm:linux', ['package:linux'], (done) => {
```
MarshallOfSound: This should sequentially run `rpm:linux:32` and `rpm:linux:64`

gulpfile.babel.js
MarshallOfSound: I.e. this `make:linux` should sequentially run `rpm:linux` and `deb:linux`

jostrander: Well, I can throw it in there with run sequence in this PR if you want. I had it in the last PR but it was overkill/not enough in different spots.

MarshallOfSound: Explain
> not enough in different spots

?

jostrander: Since run sequence runs inside of a task, it doesn't properly merge dependencies and therefore I was getting rebuilds in the 4-8 times just by running `make:linux`.

gulpfile.babel.js
jostrander: Added the separate linux builds as suggested here and it broke the CI builds, even though it reported a success.

MarshallOfSound:
Sorry, I thoroughly brutalized this PR with conflicts.  I took the base logic and pushed it to master after some refactoring

PR #984, opened 2016-04-25
Add history navigation (back/forward, like in browser)

dennislysenko:
Didn't see an immediately noticeable option for this so I went ahead and added it for personal use. Feel free to merge if desired :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules/
 build/
 dist/
 .settings.json
+.idea
--- a/src/inject/core/index.js
+++ b/src/inject/core/index.js
@@ -3,6 +3,7 @@ import './uninstall';
 import './updateAvailable';
 import './appDetails';
 import './lyrics';
+import './navigate';
 
 // Anything that requires the DOM needs to go here as of Electron 0.36.6
 document.addEventListener('DOMContentLoaded', () => {
--- a/src/inject/core/navigate.js
+++ b/src/inject/core/navigate.js
@@ -1,0 +1,15 @@
+/**
+ * Created by dennis on 4/25/16.
+ */
+
+window.addEventListener('load', () => {
+  if (!window.$) return;
+
+  Emitter.on('navigate:back', () => {
+    document.querySelector('webview').executeJavaScript('window.history.back();');
+  });
+
+  Emitter.on('navigate:forward', () => {
+    document.querySelector('webview').executeJavaScript('window.history.forward();');
+  });
+});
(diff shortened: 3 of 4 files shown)


MarshallOfSound:
@dennislysenko There are already a wide varierty of ways to navigate forward / backward

* Alt + Left/Right
* Backspace
* Mouse Button 4 + 5
* 3 Finger swipe (OS X)
* UI "back" button

From my understanding of things the "Alt + Left/Right" shortcuts are the browser forward/backward standards so I think we should stick with those.

dennislysenko:
@MarshallOfSound thanks for informing. Not sure which platform you are on, but Alt+Left/Right is not the browser standard on OSX. It is cmd+left/right. Perhaps you could include this as an OS-specific default?

dennislysenko:
Also, if you inject the UI back button manually, in my opinion it would be easier to find if it were to the left of the search icon, or somewhere more to the left. Just a thought, as I had no thought to even look there :)

MarshallOfSound:
@dennislysenko That's interesting, the `Alt + Left/Right` is standard on Linux and Windows.  It's strange that OS X would implement it differently (Round of applause for apple not following standards).

If you changed this PR to add a check to `e.metaKey` [here](https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/fd3ffe17a2228b929c7307bf0a00eedfcedab199/src/inject/GPMInject/interface/customUI.js#L187) I would accept it.

Something like

```js
      || (e.which === 37 && (e.altKey || e.metaKey) && document.activeElement.value === undefined)) {
      attemptBack();
    } else if (e.which === 39 && (e.altKey || e.metaKey) && document.activeElement.value === undefined) {
```

Just check that is works and PR it here. :+1:

dennislysenko:
I tried and... well, I only wish it were that simple. Unfortunately, as detailed [here](http://stackoverflow.com/questions/3902635/how-does-one-capture-a-macs-command-key-via-javascript), the cmd key is not processed as a modifier, for whatever boneheaded reason, but as a separate keypress. e.metaKey does not ever appear to be true, as the command and the arrow register as two separate keypresses.

dennislysenko:
In fact, the arrow key only registers as `keydown`, not `keyup`.

jostrander:
Just so you don't get too far into your research, @marshallofsound already
has all of that figured out, about the meta key not working on keyup etc.
Just trying to find a solution. Not sure which issue he has that on or if I
read it in gitter.

dennislysenko:
@jostrander thanks for the update. I actually do have it working perfectly now, so if you wish, I can clean up the code to fit the pre-commit hooks and submit a pull request. Let me know, @MarshallOfSound :)

MarshallOfSound:
@dennislysenko PR it and I'll take a look :+1:

PR #990, opened 2016-04-27
Musixmatch lyrics support

MCManuelLP:


src/main/features/core/lyrics/source_musiXmatch.js
```diff
@@ -0,0 +1,42 @@
...
+
+const decoder = new Entities();
+
+const attemptMusiXmatch = (path) => {
+  return new Promise((resolve, reject) => {
+    fetch(`https://www.musixmatch.com/search/${path}`) // the lyrics urls aren't always very straightforward...
+      .then((data) => data.text())
+      .then((searchHtml) => {
+        const searchReturn = JSON.parse(/__mxmProps = ({.+})<\/script>/.exec(searchHtml)[1]);
+        fetch(searchReturn.allResults.bestMatch.shareURI)
```
MarshallOfSound: What happens here if there is no `bestMatch`

MCManuelLP: Then it catches the error and rejects the Promise

MarshallOfSound: Seems a bit dodgy to go down the "it didn't work let the promise go to the catch method" route.  It seems akin to wrapping your entire program in a `try { ... } catch` block.  Although in this case it works and is semi appropriate, it seems bad practice in general.

src/main/features/core/lyrics/source_musiXmatch.js
```diff
@@ -0,0 +1,42 @@
...
+            lyrics = decoder.decode(lyrics);
+            lyrics = xss(lyrics, {
+              whiteList: { br: [], i: [], b: [], strong: [], em: [] },
+              stripIgnoreTag: true,
+              stripIgnoreTagBody: ['script'],
+            });
+            resolve(lyrics);
+          })
+          .catch(() => {
+            reject();
```
MarshallOfSound: This can be changed to  

```js
.catch(reject);
```

src/main/features/core/lyrics/source_musiXmatch.js
```diff
@@ -0,0 +1,42 @@
...
+          });
+      })
+      .catch(() => {
+        reject();
```
MarshallOfSound: This can be changed to  

```js
.catch(reject);
```

src/main/features/core/lyrics/source_musiXmatch.js
MCManuelLP: Well, fetch(undefined) automatically triggers the catch block afaik so I don't think that wrapping the whole thing in `try { ... } catch` would do anything

MarshallOfSound: @MCManuelLP I'm saying that letting a promise fail due to a syntax / execution error is akin to (same as) wrapping your entire code in a try catch.  Not that you should do that, in fact, the complete opposite.  I'm saying you should never wrap your code in a try catch and let it fail.  Rather check for failures / potential errors specifically at teach step.

However in this case I believe it is OK as any failure inside this method should simply reject the promise and by nature of JS errors that is what happens :+1:

PR #1000, opened 2016-04-30
Expose package:linux:32 and package:linux:64

BrianAllred:
Having deb and rpm for different architectures is great, but it then requires that the user have those package systems installed, which is not ideal for users of distributions that don't use debs or rpms. Being able to package for different architectures would prevent that issue.

Edit: Grammar.


MarshallOfSound:
LGTM :+1: 

Just waiting on CI

PR #1001, opened 2016-05-01
Add cmd-left/right history navigation on OSX

dennislysenko:
Reworked based on comments on #984.

### DIFF
--- a/src/inject/GPMInject/interface/customUI.js
+++ b/src/inject/GPMInject/interface/customUI.js
@@ -189,14 +189,13 @@ function installBackButton() {
    separate keypress, which also eats the keyup event of the key it
    is modifying.
    */
-  let isHoldingCmdKey = false;
+  const holdingKeyMap = {};
 
   window.addEventListener('keydown', (e) => {
-    if (e.which === 91 || e.which === 93) {
-      isHoldingCmdKey = true;
-    }
+    holdingKeyMap[e.which] = true;
 
-    if (isHoldingCmdKey && document.activeElement.value === undefined) {
+    if ((holdingKeyMap[91] === true || holdingKeyMap[93] === true)
+      && document.activeElement.value === undefined) {
       if (e.which === 37) {
         attemptBack();
       } else if (e.which === 39) {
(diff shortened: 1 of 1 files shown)


src/inject/GPMInject/interface/customUI.js
```diff
@@ -182,13 +182,40 @@ function installBackButton() {
...
   window.addEventListener('keyup', (e) => {
     if ((e.which === 8 && document.activeElement.value === undefined)
       || (e.which === 37 && e.altKey && document.activeElement.value === undefined)) {
       attemptBack();
     } else if (e.which === 39 && e.altKey && document.activeElement.value === undefined) {
       attemptForward();
     }
+
+    if (e.which === 91 || e.which === 93) {
+      isHoldingCmdKey = false;
```
MarshallOfSound: Purely because I enjoy being nitpicky.

Isn't it feasibly possible that a user has 2 "command" keys on their keyboard (Left / Right).  In which case with this logic.

Start = false
Down Left = true
Down Right = true
Up Right = false

Even though you still are holding a command key down we thing they aren't.  Maybe store two vars for left and right or handle the logic a bit differently

dennislysenko: Yep, this is possible. See the commit I just pushed, 9b9f580. Just tested the case of holding down left, holding down right, letting go left, so still holding right and pressing left and it navigated properly.

src/inject/GPMInject/interface/customUI.js
```diff
@@ -182,13 +182,37 @@ function installBackButton() {
...
+   However cmd does not show up as a modifier key, but rather as a
+   separate keypress, which also eats the keyup event of the key it
+   is modifying.
+   */
+  const holdingKeyMap = {};
+
+  window.addEventListener('keydown', (e) => {
+    holdingKeyMap[e.which] = true;
+
+    if ((holdingKeyMap[91] === true || holdingKeyMap[93] === true)
```
MarshallOfSound: This can be simplified to 

```js
if ((holdingKeyMap[91] || holdingKeyMap[93])
```

MarshallOfSound:
One small comment and some merge conflicts that have to be resolved.  Other than that LGTM

shakeelmohamed:
Can I help finish this up?

MarshallOfSound:
@shakeelmohamed This is so far behind master I don't even think it could merge if I kicked it 😆 

Might be better to just start from scratch 😆

shakeelmohamed:
@MarshallOfSound good to know 😄

PR #1086, opened 2016-05-17
Update Installation_On_Debian.md

jostrander:
Dekstop yo. :+1:


MarshallOfSound:
So, I totally copy and pasted that.  From Gitter.  Where you wrote it.

LOL

Issue #883, opened 2016-04-03
Installation Failed on Fedora 23


DanielWEWO:
**OS:**
Fedora 23 
4.4.6-300.fc23.x86_64

**Issue Descriptions:**
Installation through software center fails with error message "Sorry, this did not work; Installation of An electron wrapper for Google Play Music failed."

**Steps to Reproduce:**

Download 64 bit .rpm from https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/releases, double click to attempt to install with software center

Would be awesome to be able to use this! This works great on my laptop on Ubuntu.

Also, attempting to install with dnf

`sudo dnf install google-play-music-desktop-player-3.1.0.amd64.rpm 
[sudo] password for dlynch: 
Error: package google-play-music-desktop-player-3.1.0-1.amd64 is not installable
(try to add '--allowerasing' to command line to replace conflicting packages)
[dlynch@localhost Downloads]$ sudo dnf install google-play-music-desktop-player-3.1.0.amd64.rpm --allowerasing
Error: package google-play-music-desktop-player-3.1.0-1.amd64 is not installable
`

MarshallOfSound:
Did you try installing with the `--allowerasing` flag as the error suggests?

I think I've seen this before and that flag fixed it.

DanielWEWO:
Yeah, if you look closer you can see that I tried it again with that flag, sorry for the bad formatting.

`[dlynch@localhost Downloads]$ sudo dnf install google-play-music-desktop-player-3.1.0.amd64.rpm --allowerasing Error: package google-play-music-desktop-player-3.1.0-1.amd64 is not installable`

jostrander:
I mean quite honestly it's a pretty vague error, what's it conflicting with? FWIW, triple back ticks on the line before and after will avoid line breaks. 

```
[dlynch@localhost Downloads]$ sudo dnf install google-play-music-desktop-player-3.1.0.amd64.rpm --allowerasing Error: package google-play-music-desktop-player-3.1.0-1.amd64 is not installable
```

DanielWEWO:
Thanks. Yeah, it is really vague and I'm not sure what's conflicting it's actually a relatively new Fedora installation. I tried installing with the verbose option and it didn't give any more information.

jostrander:
Can you try running with `--verbose` ?

DanielWEWO:
[...]

PR #1123, opened 2016-05-27
Use x86_64 architechure for fedora/redhat rpms

tiwillia:
Fedora/redhat based 64-bit systems use the architechure 'x86_64'.
Packages of the architechure amd64 will be unable to be installed using
default package management tools, such as dnf or yum.

This should resolve https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/issues/883

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -257,9 +257,14 @@ const generateGulpLinuxDistroTask = (prefix, name, arch) => {
       categories: ['AudioVideo', 'Audio'],
     };
 
+    let pkg_arch = 'i386';
+    if (arch === '64') {
+      pkg_arch = (prefix === 'rpm' ? 'x86_64' : 'amd64');
+    }
+
     tool(_.extend({}, defaults, {
       src: `dist/${packageJSON.productName}-linux-${arch === '32' ? 'ia32' : 'x64'}`,
-      arch: arch === '32' ? 'i386' : 'amd64',
+      arch: pkg_arch,
     }), (err) => {
       console.log(`${arch}bit ${prefix} package built`); // eslint-disable-line
       if (err) return done(err);


jostrander:
Tests are obviously failing for reasons other than this PR, however I'm kind of torn as to just have it be formatted as just:

```js
    arch === '32' ? 'i386' : prefix === 'rpm' ? 'x86_64' : 'amd64'
```

where the original line was to avoid the lines and variable above.

tiwillia:
Sorry, I don't usually write nodejs. I can make that change.

zeroepoch:
I just ran into this problem as well.  `dnf install ./google-play-music-desktop-player-3.2.4.amd64.rpm` doesn't work with the same random error mentioned in #883. It works with `rpm -Uvh google-play-music-desktop-player-3.2.4.amd64.rpm`. It's because of this amd64 vs x86_64 issue most likely.  Somehow though I think dnf worked on my laptop but fails like I would expect on my desktop.  Could be I used rpm and just forgot.

gulpfile.babel.js
```diff
@@ -257,9 +257,18 @@ const generateGulpLinuxDistroTask = (prefix, name, arch) => {
       categories: ['AudioVideo', 'Audio'],
     };
 
+    var pkg_arch = ''
+    if (arch === '32') {
+      pkg_arch = 'i386';
+    } else if (prefix === 'rpm') {
+      pkg_arch = 'x86_64';
+    } else {
```
MarshallOfSound: Purely because I like being picky and having my own JS style forced onto others :laughing:   I feel this would make more logical sense as 

```js
let pkg_arch = 'i386';
if (arch === '64') {
    pkg_arch = (prefix === 'rpm' ? 'x86_64' : 'amd64');
}
```

Also pay attention to the use of `let` instead of `var`.  It's an ES6 thing :+1:

MarshallOfSound:
@jostrander FYI nested ternary expressions violate the linting rules :+1: 

@tiwillia Once small comment to fix up then this can go to master

tiwillia:
@MarshallOfSound, I made your recommended change, let me know if there is anything else needed.

Issue #972, opened 2016-04-20
Lyrics move too slow or too fast


Double44:
I use the Lyrics feature a ton, I just love it. The service you are using to retrieve them is very good. However I find that on most songs, the lyrics aren't always paced very well. Often the page scrolls way too slow, and other songs it can move too fast. I really like the automatic scrolling, but a manual scrollbar could be useful too, including the use of up/down keys.

**OS:**
Windows 10

**Issue Descriptions:**
Lyrics scroll too slow or too fast

**Steps to Reproduce:**
Play song, click 3-dots, Show Lyrics (Beta)

MarshallOfSound:
@Double44 There is a keyboard shortcut to access the lyrics "Cntrl + Shift + L"

Can you name a few songs that the auto scroll doesn't quite work for.  I tested on a stupid quantity of songs and the speeds / timing I ended up using worked for most of them

Double44:
In Stereo - Fort Minor
Petrified - Fort Minor
Clint Eastwood - Gorillaz
Welcome to the World of the Plastic Beach - Gorillaz
Super-Charger Heaven - White Zombie
Black Sunshine - White Zombie

The Petrified lyrics are slightly messed up, which might account for the extra speediness.

jostrander:
Unfortunately without timestamped lyrics the scrolling will never work for absolutely all songs. Maybe have the auto scroll be a setting that can be checked on the lyrics screen?

wirespot:
Or consider showing lyrics with manual scroll, and let the user look at whatever part they want. The autoscroll misbehaves in lots of cases, and becomes unusable if you use "prev" to restart the song or skip inside it. IMHO the autoscroll looks like more trouble than it's worth.

andrewjmetzger:
+1 for removing autoscroll. No need to do that, we have scroll wheels and trackpads! After that, lyrics integration would be good to go.

Lettuceman44:
Listening to True Survivor - David Hasselhoff, the timing is also off. I would agree auto scroll isn't needed, or at least make it an option.

Double44:
Yes I think the autoscroll should be an option.. I wouldn't completely do away with it because when it works, it's great. If the autoscroll could allow manual input and continue to scroll after that input, I think this would be a good workaround.

eRadicant:
Also, lyric scroll position isn't adjusted if you seek forwards or backwards in the song.

[...]

PR #1124, opened 2016-05-29
Temporarily comment out autoscroll for 3.3.0

jostrander:
Until a solution is figured out for auto scroll, lets disable it. Alternatives are welcome, but the majority of the vote on issue #972 was to just nix it.


MarshallOfSound:
This one I'm going to have to deny

* Commenting out code is not the way to go here
* Making scrolling optional is (Just make a settings option)

PR #1131, opened 2016-05-31
Overhauled lyrics screen

MCManuelLP:
Light theme and pink:
light-theme
Dark theme and teal:
dark-theme
No theme:
no-theme

### DIFF
--- a/src/inject/generic/core/lyrics.js
+++ b/src/inject/generic/core/lyrics.js
@@ -28,7 +28,7 @@ window.addEventListener('load', () => {
   };
   // Handle time progression of a song
   const timeHandler = (timeObj) => {
-    $('#lyrics_back > #lyrics_progress > #lyrics_bar').width(`${timeObj.current / timeObj.total * 100}%`);
+    $('#lyrics_back > #lyrics_progress > #lyrics_bar').width(`${timeObj.current / (timeObj.total+1) * 100}%`);
     let jumped = false;
     if (timeObj.current < jumpDetect || timeObj.current - jumpDetect > 1000) {
       animate = true;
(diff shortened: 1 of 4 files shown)


src/inject/generic/core/lyrics.js
```diff
@@ -28,6 +28,7 @@ window.addEventListener('load', () => {
   };
   // Handle time progression of a song
   const timeHandler = (timeObj) => {
+    $('#lyrics_back > #lyrics_progress > #lyrics_bar').width(`${timeObj.current / timeObj.total * 100}%`);
```
MarshallOfSound: It is entirely possible for `timeObj.total` to be `0`.  That needs to be handled somehow

src/inject/generic/windowThemeHandler.js
```diff
@@ -32,7 +32,11 @@ if (window.$ && window.$.ajax) {
     const border = `[theme][light] .window-border{border-color:${color}}`;
     const titleBar = `[theme][light] .title-bar{background:${color}}`;
     const header = `[theme][light] .dialog .window-title{background:${color}}`;
-    style.html(border + titleBar + header);
+    const lyricsProgress = `\
```
MarshallOfSound: These slashes are not required when using ES6 strings

MCManuelLP:
I'm not sure if the +1 solution is the best but it's the easiest :laughing: (and shouldn't impact it noticeably)

src/inject/generic/core/lyrics.js
```diff
@@ -28,6 +28,7 @@ window.addEventListener('load', () => {
   };
   // Handle time progression of a song
   const timeHandler = (timeObj) => {
+    $('#lyrics_back > #lyrics_progress > #lyrics_bar').width(`${timeObj.current / (timeObj.total + 1) * 100}%`);
```
MarshallOfSound: This would be better

```
const { current, total } = timeObj;
`${(total === 0 ? 0 : current / total) * 100}%`
```

PR #1169, opened 2016-06-09
Allow overriding WebSocketAPI listening interface

thsnr:
Currently, the created WebSocket server listens on all interfaces, which is a security concern. Listen only on the local interface by default.

### DIFF
--- a/src/main/features/core/websocketAPI.js
+++ b/src/main/features/core/websocketAPI.js
@@ -76,7 +76,10 @@ const enableAPI = () => {
     });
     Emitter.sendToWindowsOfName('main', 'openport:request');
   }
-  server = new WebSocketServer({ port: global.API_PORT || process['env'].GPMDP_API_PORT || 5672 }, () => { // eslint-disable-line
+  server = new WebSocketServer({
+    host: process['env'].GPMDP_API_HOST || '0.0.0.0', // eslint-disable-line
+    port: global.API_PORT || process['env'].GPMDP_API_PORT || 5672, // eslint-disable-line
+  }, () => {
     if (ad) {
       ad.stop();
       ad = null;


MarshallOfSound:
Regarding your PR
------------------------
> Currently, the created WebSocket server listens on all interfaces

This is by design, we want to be able to use the PlaybackAPI on the local network

> which is a security concern

The WebSocketAPI is off by default, if you turn it on you should be aware that you are enabled remote access to your player.  The message in the settings window informs you of this.

> Listen only on the local interface by default.

This would render the API uselss


Regarding Security
------------------------
The WebSocketAPI provides read only access by default, if an applications wants to control the player it has to go through an authentication sequence which involves the user manually inputting a random 4 digit number which is generated inside the player.

More information on this authentication process can be found in the [WebSocketAPI Docs](https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/master/docs/PlaybackAPI_WebSocket.md#be-polite)

thsnr:
Ok, your reasoning is sound. However, how about incorporating the patch, but putting 0.0.0.0 as the default? This way everything continues to work as before, but someone that does not wish GPMDP to listen on all interfaces can control it using the GPMDP_API_HOST environment variable.

Even though there is the 4-digit PIN security measure, it might be undesirable to provide even read-only access to the local network, while still wanting control from the local machine (or only from a specific interface). This could also be achieved by configuring a firewall, but it is simpler to just not listen on any extra interfaces.

Of course, I can always only patch it locally, but maybe someone else wishes for this feature also.

MarshallOfSound:
@thsnr That sounds like a fair enough request.

If you change the default to be `'0.0.0.0'` then I would accept the PR.  Please ensure that you use single quotes for the string (It's a linting thing)

src/main/features/core/websocketAPI.js
```diff
@@ -76,7 +76,10 @@ const enableAPI = () => {
     });
     Emitter.sendToWindowsOfName('main', 'openport:request');
   }
-  server = new WebSocketServer({ port: global.API_PORT || process['env'].GPMDP_API_PORT || 5672 }, () => { // eslint-disable-line
+  server = new WebSocketServer({
+    host: global.API_HOST || process.env.GPMDP_API_HOST || '0.0.0.0',
+    port: global.API_PORT || process.env.GPMDP_API_PORT || 5672,
```
MarshallOfSound: This NEEDS to be

```
process['env'].GPMDP_API_HOST
process['env'].GPMDP_API_PORT
```

You've got to be careful when changing code in a project where you don't understand the build tooling.  We inline replace `process.env.XXXXX` with the **build machines** environment variables.  If we want to use the client's environment variables we have to use this syntax.

See https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/master/gulpfile.babel.js#L156

src/main/features/core/websocketAPI.js
```diff
@@ -76,7 +76,10 @@ const enableAPI = () => {
     });
     Emitter.sendToWindowsOfName('main', 'openport:request');
   }
-  server = new WebSocketServer({ port: global.API_PORT || process['env'].GPMDP_API_PORT || 5672 }, () => { // eslint-disable-line
+  server = new WebSocketServer({
+    host: global.API_HOST || process.env.GPMDP_API_HOST || '0.0.0.0',
```
MarshallOfSound:
```js
global.API_HOST
```

Is not required, we only use `global.API_PORT` for testing purposes.

thsnr:
> You've got to be careful when changing code in a project where you don't understand the build tooling. We inline replace process.env.XXXXX with the build machines environment variables. If we want to use the client's environment variables we have to use this syntax.

Ah, okay. The linter was whining at me, but this explains why that line was disabled. My bad, fixed now.

> `global.API_HOST`
> 
> Is not required, we only use global.API_PORT for testing purposes.

Removed.

PR #1211, opened 2016-06-19
Add support for additional hotkeys

caseif:
This PR is straightforward and simple, adding support for hotkeys other than letters and arrow keys. A list of keys this PR supports is below.

- Digit keys
- Standard function keys (`F1`–`F12`)
- A variety of symbols (``-=[];',./`\``)
- `Insert`, `Delete`, `Home`, `End`, `PageUp`, `PageDown`

### DIFF
--- a/src/inject/settings/hotkeys.js
+++ b/src/inject/settings/hotkeys.js
@@ -9,14 +9,35 @@ const MODIFIER_KEYS = {
   18: 'Alt',
 };
 
-const ACTION_KEYS = _.transform(_.range(26), (final, current) => {
+const ACTION_KEYS = _.transform(_.range(26), (final, current) => { // letters
   final[current + 65] = String.fromCharCode(current + 65); // eslint-disable-line
+}, _.transform(_.range(10), (final, current) => { // digits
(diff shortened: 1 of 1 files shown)


src/inject/settings/hotkeys.js
```diff
@@ -9,13 +9,36 @@ const MODIFIER_KEYS = {
   18: 'Alt',
 };
 
-const ACTION_KEYS = _.transform(_.range(26), (final, current) => {
+const ACTION_KEYS = _.transform(_.range(26), (final, current) => { // letters
   final[current + 65] = String.fromCharCode(current + 65); // eslint-disable-line
+  if (current < 12) {
```
MarshallOfSound: Tagging the F keys and number keys into this function is really strange and I don't like it 😆 

Can you separate them into their own transforms

caseif: That was how it was written in the original commit, but I thought this format was decidedly more readable. I can change it back, though.

MarshallOfSound:
LGTM 👍

PR #1214, opened 2016-06-20
Add option to disable auto scroll

MCManuelLP:
Solves #972


src/_locales/pl-PL.json
```diff
@@ -81,6 +81,8 @@
 
   "settings-option-auto-launch": "Włącz aplikację podczas uruchamiania systemu.",
 
+  "settings-option-scroll-lyrics": "Scroll lyrics automatically while playing",
```
MCManuelLP: I know, will remove that, but later, for when you find some more stuff to correct :laughing:

MCManuelLP:
Plus some of the comments can go... They just help me stay on track :smile:

src/inject/generic/core/lyrics.js
```diff
@@ -18,7 +20,7 @@ window.addEventListener('load', () => {
     } else {
       clearTimeout(noLyricsTimer);
       const lyricsHTML = lyrics.replace(/\n/g, '<br />');
-      $('#lyrics').html(`<p>${lyricsHTML}</p>`);
+      $('#lyrics').html(`<p scroll="${Settings.get('scrollLyrics', true)}">${lyricsHTML}</p>`);
```
MarshallOfSound: Using custom attributes should be `data-scroll=`

Thought why are we using HTML attributes anyway?

MCManuelLP: Hmm, easiest way (I know) to have a css toggle of the scrollbar...

MarshallOfSound: Fair enough, but year, please change to `data-scroll` it is more semantically correct.

src/assets/less/main_window.less
```diff
@@ -194,9 +194,14 @@ html, body {
         width: 100%;
         font-size: 26px;
         line-height: 1.6em;
-        overflow: hidden; /* Hide scroll bar so that the user relies on auto scroll */
         padding: 12px 0;
       }
+      p[scroll=true] {
```
MarshallOfSound: It would make more sense to have an attribute represent a boolean value

I.e. The existance of `data-scroll` = true, if `data-scroll` is not set, then false.

In JS you would use `.setAttribute` and `.removeAttribute`

src/inject/generic/core/lyrics.js
```diff
@@ -52,9 +54,28 @@ window.addEventListener('load', () => {
     animate = false;
   };
 
+  const scrollHandler = (state) => {
```
MarshallOfSound: This isn't so much a `scrollHandler` as a `scrollSettingsHandler`, they are semantically different

src/inject/generic/core/lyrics.js
MCManuelLP: Will do

src/inject/generic/core/lyrics.js
```diff
@@ -52,9 +54,28 @@ window.addEventListener('load', () => {
...
+    const lyricsP = $('#lyrics p');
+    if (state) { // enable auto scroll
+      lyricsP.attr('scroll', true);
+      animate = true;
+      timeHandler(remote.getGlobal('PlaybackAPI').data.time);
```
MarshallOfSound: We should call `clearTimeout(animationTimer)` before this line

IPC events are asynchronous, we don't want to run into any problems 😆

src/inject/settings/general.js
```diff
@@ -46,3 +46,17 @@ if (Settings.get('nativeFrame')) {
 } else {
   $('#native-frame').removeAttr('checked');
 }
+
+$('#scroll-lyrics').change((e) => {
+  Emitter.fire('settings:set', {
+    key: 'scrollLyrics',
+    value: $(e.currentTarget).is(':checked'),
+  });
+  Emitter.fireAtMain('settings:set:scrollLyrics', $(e.currentTarget).is(':checked'));
```
MarshallOfSound: This should be handled in the handler for `settings:set`.  See how we handle other events that require passthrough here.

https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/master/src/main/features/core/desktopSettings.js#L67

src/inject/generic/core/lyrics.js
MCManuelLP: Thinking about it again, technically we really don't need this one do we?

MarshallOfSound: Don't need what?

MCManuelLP: The extra call to timeHandler

MarshallOfSound: Remove it then 😆 

If it works without it 😆

MCManuelLP:
I have to rebase don't I ...

src/assets/less/main_window.less
```diff
@@ -194,8 +194,11 @@ html, body {
         width: 100%;
         font-size: 26px;
         line-height: 1.6em;
-        overflow: hidden; /* Hide scroll bar so that the user relies on auto scroll */
         padding: 12px 0;
+        overflow: overlay;
```
MarshallOfSound: > overlay: non-standard value only supported by the WebKit-browsers that does the same as auto.

Please use `auto` 😆

MCManuelLP: When I used `auto` it messed up the layout and it moves the text... But sure...

MCManuelLP: Okay, whatever made it behave weirdly earlier doesn't seem to be a problem anymore, ffix incoming

src/assets/less/main_window.less
```diff
@@ -194,8 +194,11 @@ html, body {
         width: 100%;
         font-size: 26px;
         line-height: 1.6em;
-        overflow: hidden; /* Hide scroll bar so that the user relies on auto scroll */
         padding: 12px 0;
+        overflow: auto; //overlay;
```
MarshallOfSound: "//overlay;" this comment is not required

MCManuelLP: I should totally look over the changes before I push...

MarshallOfSound:
LGTM 👍 

Just waiting for CI

PR #1249, opened 2016-06-28
Add installers, admin permission checks, better directory handling

JoshuaGarrison27:
Added Node and Git
Admin checks to help the user understand why the script might not work
Better handling of the current directory

### DIFF
--- a/vendor/setup_window.ps1
+++ b/vendor/setup_window.ps1
@@ -8,14 +8,15 @@ function Test-Administrator
 #Check for Node install
 function Node-Checker
 {
-    Try
-    {
-        return node -v
-    }
-    Catch
-    {
-        return ""
-    }
+    Try { return node -v }
+    Catch { return "" }
+}
+
+#Check for Git install
+function Git-Checker
+{
+    Try {return git --version}
+    Catch {return ""}
 }
 
 #Perform test to see if user as an administrator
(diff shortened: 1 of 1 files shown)


vendor/setup_window.ps1
```diff
@@ -12,6 +54,15 @@ $bonjour_sdk = "$storageDir\bonjoursdksetup.exe"
 $python_remote = "https://www.python.org/ftp/python/2.7.10/python-2.7.10.amd64.msi"
 $python = "$storageDir\python.msi"
 
+Write-Host " "
+Write-Host "Downloading Node: $node_installer"
+$webclient.DownloadFile($node_installer_remote, $node_installer)
```
MarshallOfSound: Installing this could be bad as users could already have specific node versions installed.  We need to somehow ask the user for input.  I'll look into this in PowerShell

JoshuaGarrison27:
Good point. I have committed a check for Node. I thought about applying it to most of the installs but I thought that might be overkill.

MarshallOfSound:
@JoshuaGarrison27 The node check looks good 👍 

However I'm not keen on the git install, I just tried it and it overwrote my current git install (including all my dot files, global configs and such).  We should definitely check for git if the user has git installed already or not before installing.

(Don't ask, just only install if missing)

MarshallOfSound:
LGTM 👍 

Thanks for this 😆

JoshuaGarrison27:
You're welcome. I may write it better soon to try and clean up the messy look of the script.

PR #1339, opened 2016-07-21
reimplement last.fm scrobbling

petuhovskiy:
watch playback time and scrobble when it reaches 60% of song duration

### DIFF
--- a/src/inject/GPMWebView/playback/handler.js
+++ b/src/inject/GPMWebView/playback/handler.js
@@ -15,12 +15,14 @@ window.wait(() => {
     }
   });
 
-  let lastScrobble = {};
-  let lastScrobbleTime = 0;
   let currentSong;
-
+  let playTime = 0;
+  let lastPosition = 0;
+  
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
The logic you have changed to internally makes little sense to me.

Can you please explain it either on GitHub or in comments.  For instance the whole `progress > 0 && progress < 2000` logic makes literally no sense 😕

MarshallOfSound:
And FWIW the song should scrobble when it reaches `50%` which is what it already does so I'm not sure what this PR actually changes???

petuhovskiy:
'change:playback-time' event fires every ~200ms, so it handles manual position change, when progress too big or negative

petuhovskiy:
currently you can alternately click on the start and end positions of song every 10 seconds and it will increase scroblles count, what is not supposed to be

MarshallOfSound:
I tested this out with other players that support last.fm and the current behavior appears to be conistent with their implementation so I'm going to close this as invalid.

Thanks for the PR though 👍

Issue #1492, opened 2016-08-19
Remember full-screen mode setting on OS X


amancevice:
**OS:** 

OS X

**Issue Descriptions:**

When using this app, I often keep it in full screen mode as its own desktop space. Some OS X apps have the ability to remember this setting when starting up ([Atom](https://atom.io/) is an example). It would be nice to add that capability to this app so that it starts up in the same mode it was in when it quit.

**Steps to Reproduce:**

Open app, go full screen with `⌃⌘F`, quit app with `⌘Q`, open app again -- app is not full screen

MarshallOfSound:
This was a design decision I made very early on, mainly because as a user I don't expect to open a media player and it to instantly go full screen.

That isn't normal, standard or done anywhere else I believe 👍 

Marking as `wontfix` due to design decision.  If you disagree I welcome alternative viewpoints but you'd have to provide examples of media apps doing it already in the wild before I would consider it.

amancevice:
I do disagree, I think it's reasonable to assume that a user would like the state of the app to remain consistent. To be clear, I'm *not* saying *every time* the app starts it should be in full screen mode. Rather, that it should start in the mode it was in when it was last open.

Many OS X apps do this. The terminal app, Chrome, and Atom are examples. If you open Chrome on OS X, make it full screen, quit, and re-open it will open as a full screen app again. I don't think the fact that this app is a media app is relevant in this context -- it's behavior that I think could apply to all OS X apps.

amancevice:
I should clarify -- the apps that I mentioned (Chrome, Terminal, Atom) behave this way if you adjust your OS X settings in `System Preferences > General > Uncheck "Close windows when quitting an application"`, which I don't believe is the default state of that checkbox.

Apologies; I've had that setting disabled for so long I forgot about it.

So perhaps what I am asking for is that GPMDP not close its window on quit so that OS X can take advantage of this setting.

amancevice:
@MarshallOfSound Any response to the above? Apologies if you've already seen it.

PR #1569, opened 2016-09-02
Remain FullScreen

amancevice:
Use electron-window-state package to remember full screen state of app on quit.
* If app if full screen on quit, it will start in full screen mode
* If app is NOT full screen on quit, it will start normally

Fixes [Issue#1492](https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/issues/1492)

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "auto-launch": "^4.0.0",
     "electron": "1.3.4",
     "electron-chromecast": "^1.0.14",
+    "electron-window-state": "^3.0.4",
     "gmusic-mini-player.js": "^2.0.6",
     "gmusic-theme.js": "^2.0.0",
     "gmusic-ui.js": "^2.1.0",
--- a/src/index.js
+++ b/src/index.js
@@ -33,6 +33,8 @@ updateShortcuts();
   // be closed automatically when the JavaScript object is garbage collected.
   let mainWindow = null;
 
+  const windowStateKeeper = require('electron-window-state');
+
   // DEV: Make the app single instance
   const shouldQuit = app.makeSingleInstance(() => {
     if (mainWindow) {
@@ -112,6 +114,12 @@ updateShortcuts();
   // initialization and is ready to create browser windows.
   app.on('ready', () => {
     mainWindow = new BrowserWindow(generateBrowserConfig());
+
+    // Set full screen if previous state was full screen
+    const mainWindowState = windowStateKeeper();
+    mainWindow.setFullScreen(mainWindowState.isFullScreen);
+    mainWindowState.manage(mainWindow);
+
     global.mainWindowID = WindowManager.add(mainWindow, 'main');
 
     const position = Settings.get('position');


MarshallOfSound:
@amancevice I don't like pulling in modules to do something as simple as persist a boolean value.

We are literally pulling in an entire dependency to just remember `true` and `false`

Also, I believe I made my stance relatively clear in issue #1492.  I see there are extra comments on that issue that I have missed so I will respond to them here.

> I do disagree, I think it's reasonable to assume that a user would like the state of the app to remain consistent. To be clear, I'm not saying every time the app starts it should be in full screen mode. Rather, that it should start in the mode it was in when it was last open.

To put this idea in perspective, let's compare the request to the actions of iTunes, VLC and Quicktime as they are "media applications".  I tested all of the above 3 applications and on my system none of them persisted fullscreen mode.  I also tested other Electron applications (Atom and Quick Start), neither persisted the fullscreen mode.

> So perhaps what I am asking for is that GPMDP not close its window on quit so that OS X can take advantage of this setting.

This is impossible in Electron applications.  The windows are the core UI processes so if you leave them open the app remains open.

amancevice:
@MarshallOfSound I can see your point about pulling in a dependency without careful consideration; this was my first experience with electron so I appreciate your patience.

I assure you, however, that Atom does persist full screen state on startup as I use it every day. I am running OS X (El Cap) so if you are attempting to verify my claim on Windows or an older version of OS X then I have no idea Atom or any other apps behave.

I don't really understand the resistance to persisting app state across restart. It seems reasonable to me that a user expect an app to start where it left off, much like Netflix might resume a video on browser start or a car stereo resume a CD on ignition.

However, I'll drop the issue since you are so strongly opposed to it. I enjoy the app otherwise, thanks.

MarshallOfSound:
@amancevice We already persist a massive amount of the apps state across relaunches.

https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/master/src/main/features/core/persistAppState.js

We persist size, position, zoom factor and current GPM page.

My core issue with persisting fullscreen is  
1. From what I can see on macOS it is a non-default feature.  I.e. Most users won't expect it
2. On all other platforms it is definitely not a thing.  No one would expect it
3. Even on macOS it is a non-standard feature.  I.e. Not all apps implement it

Considering the userbase the app targets and the fact the feature is supposed to be both non-standard and non-default I don't consider it a viable feature to implement 👍

amancevice:
@MarshallOfSound I think we are talking past each other or are using different operating systems. I wonder if you understanding my use of "full screen" to mean "maximize", which is not what I intended.

In response to your core issues:

1. I don't know what you mean by non-default feature. It is a well-described feature in the [Mac App Programming Guide](https://developer.apple.com/library/mac/documentation/General/Conceptual/MOSXAppProgrammingGuide/FullScreenApp/FullScreenApp.html). Notably the line `Media-rich apps, on the other hand, can often benefit from full-screen presentation.`
2. I have no idea how `win.setFullScreen(true);` behaves on Windows or Linux. So you are correct to be skeptical here. I don't have access to a windows machine but I could probably test on ubuntu.
3. Again, I don't know what you mean by non-standard featue. I said originally that apps like Chrome -- the most popular browser (app?) in the world -- behave this way. You preferred to use other media apps like Quicktime, VLC player, and iTunes as comparisons. I confirmed that you are correct that Quicktime and VLC do not display this behavior, but iTunes certainly does. I tested this on my machine with and without the "Close windows when quitting app" setting enabled and the result is the same: if I am full screen when I quit, I am full screen when I open. I don't know how many average users are using Quicktime or VLC on the regular.

PR #1623, opened 2016-09-11
Implementation of Plugins

MarshallOfSound:
🎉 

Example plugin can be found here https://github.com/MarshallOfSound/gpmdp-plugin-now-playing-text-file

### DIFF
--- a/src/_locales/de.json
+++ b/src/_locales/de.json
@@ -106,6 +106,7 @@
   "title-settings-hotkeys": "Hotkeys",
   "title-settings-lastfm": "LastFM",
   "title-settings-mini": "Mini Player",
+  "title-settings-plugins": "",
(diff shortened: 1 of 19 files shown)


src/_locales/de.json
```diff
@@ -106,6 +106,7 @@
   "title-settings-hotkeys": "Hotkeys",
   "title-settings-lastfm": "LastFM",
   "title-settings-mini": "Mini Player",
+  "title-settings-plugins": "",
```
levrik: Would be just `Plugins` in German too :)

MarshallOfSound:
I've since implemented plugins in other Electron apps and have learnt a lot.  I'll start this from scratch in a new direction 👍

PR #1625, opened 2016-09-11
Added tray icons for play/pause

berryelectronics:
All references for #1275 , but didnt had the time to add the code, so i dont mind if you dont want it now.


MarshallOfSound:
@berryelectronics I updated the base branch but can you rebase onto it.

All work for `v4` should be targetted at the `v4` branch

PR #1642, opened 2016-09-15
If a WebSockets API method call returns a promise, resolve that promise before returning a result.

### LINKED ISSUE #1641: WebSocket Playback API has an empty return value for methods that invoke promises
**OS:**
Ubuntu 16.04

**Steps to Reproduce:**
   `{
      'namespace': 'search',
      'method': 'performSearch',
      'arguments': ['test search'],
      'requestID': 2
    }`
Sending the above message gives this result:
`{ namespace: 'result', requestID: 2, type: 'return', value: {} }`

GPMDP might need to make an effort to resolve promises and return through WebSockets only then.

arilotter:
Resolves #1641

### DIFF
--- a/src/inject/GPMWebView/playback/handler.js
+++ b/src/inject/GPMWebView/playback/handler.js
@@ -89,11 +89,13 @@ window.wait(() => {
       } catch (err) {
         error = err;
       }
-      Emitter.fire(`execute:gmusic:result_${cmd.requestID}`, {
-        namespace: 'result',
-        type: error ? 'error' : 'return',
-        value: error || result,
-        requestID: cmd.requestID,
+      Promise.resolve(result).then(resolved => {
+        Emitter.fire(`execute:gmusic:result_${cmd.requestID}`, {
+          namespace: 'result',
+          type: error ? 'error' : 'return',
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
RIP

Already fixed in c89b30f3b0822463bc2dc12b8bf621005db76b3f

Didn't realize you were going for the fix 😆

arilotter:
LOL, thanks for fixing it! I like your solution better anyways :+1:

MarshallOfSound:
The only real difference is mine handles promise rejection as `error` responses.

https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/commit/c89b30f3b0822463bc2dc12b8bf621005db76b3f#diff-a4cddee37a18c64bce654445aaa81e58R105

Out of interest, what are you using the API for 👍 ?

arilotter:
I'm building a dorm room automation system & automating GPMDP with voice commands from a remote device! Is there somewhere I could have a quick chat with you about another thing that might not need its own Issue?

MarshallOfSound:
@arilotter Jump on Gitter, I'm always around 👍

PR #1702, opened 2016-09-29
Convert all the things to Typescript

MarshallOfSound:
This is just a PR so I can see the changes and decide whether I want to move to typescript or stick with pure ES6


MarshallOfSound:
After looking over the PR, the changes are **REALLY** nice from a developers standpoint but from a contributors perspective this is a nightmare.

1. Linting in typescript follows a whole load of rules that contradict both `airbnb` eslint style guides and my own
2. Electron typings are not ready for the mainstream yet and throw a lot of errors which require lots of `any` casting (bad thing)
3. It increases the level of experience requried to contribute to this project which I don't want to do.

Rejecting my own PR 👍

PR #1792, opened 2016-10-18
Added English (UK)

GreenTurtwig:
Also fixed a small error in US


src/_locales/_provider.js
```diff
@@ -9,7 +9,7 @@ const localeLinks = {
   'en-AU': 'en-US',
   'en-BZ': 'en-US',
   'en-CA': 'en-US',
-  'en-GB': 'en-US',
```
MarshallOfSound: You can completely delete this line now

GreenTurtwig: Completely forgot about that. :-P

MarshallOfSound:
In general I don't like this much duplicate, pretty much all the strings are the same as `en-US` which makes adding new strings just that much more annoying.

I would suggest instead modifying `TranslationProvider` to accept js files aswell and simply use `Object.assign()` to override the few object properties you want to change for UK

jostrander:
This PR is stale, closing.

PR #1806, opened 2016-10-20
Fixing Alarm TimePicker

FunFrog-BY:
The Alarm TimePicker text field was only 256px wide, i don't know about
other languages, but in Russian caption just didn't fit. Stretched it to
the window width, now it's looking nice.

### DIFF
--- a/src/renderer/ui/components/modals/AlarmModal.js
+++ b/src/renderer/ui/components/modals/AlarmModal.js
@@ -81,7 +81,7 @@ class AlarmModal extends Component {
       >
         <TimePicker
           hintText={TranslationProvider.query('button-text-click-here-to-set-alarm')}
-          textFieldStyle={{width: '100%'}}
+          fullWidth
           onChange={this.onChange}
           {...timePickerProps}
         />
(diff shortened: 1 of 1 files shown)


jostrander:
Screen shot before/after?

FunFrog-BY:
after

FunFrog-BY:
Before:
before

src/renderer/ui/components/modals/AlarmModal.js
```diff
@@ -81,6 +81,7 @@ class AlarmModal extends Component {
       >
         <TimePicker
           hintText={TranslationProvider.query('button-text-click-here-to-set-alarm')}
+          textFieldStyle={{width: '100%'}}
```
MarshallOfSound: http://www.material-ui.com/#/components/text-field

This should be replaced with the `fullWidth` property

FunFrog-BY: Sorry, it`s my very first time with github, node, npm and of course material-ui. Can i make such minor cosmetic pull requests in future?

PR #1832, opened 2016-10-26
Added French Language

clementlamoureux:
Added French Language


src/renderer/ui/utils/constants.js
```diff
@@ -47,6 +47,7 @@ export const languageMap = {
   'nl-NL': 'Dutch',
   'pl-PL': 'Polish',
   'pt-BR': 'Portuguese',
+  'fr-FR': 'French',
```
MarshallOfSound: Please place this row in the alphabetically correct place based on it's key

MarshallOfSound:
@clementlamoureux I just updated your branch to latest master and we are now missing a translation

PR #1911, opened 2016-11-15
README dev requirements addition

chrismou:
Added reference to a couple of development packages required for compilation on Ubuntu

Let me know if you're not happy with the wording or whatever.  Also wasn't sure whether to put it in the setup instructions or dev requirements so opted for the latter :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,10 @@ Dev Requirements
 * Node.js (Recommend `4.4.x`)
 * NPM (3.x.x)
 
+To build on Ubuntu you may need to install a couple of development packages in order to compile the app:
+```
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
Maybe add a link to https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/wiki/Development-FAQ as well so other deps and platforms are there as well 👍

chrismou:
Dammit, how do I keep missing all these wiki pages?

I may as well just switch out what I added for a link to that page, and add `libexpat1-dev` to the list.  Will update the PR shortly

PR #2036, opened 2016-12-20
Add a resolver for "Go To URL" modal to follow redirects

### LINKED ISSUE #2015: Can't go to shortlinks
**OS:**

Linux (although should be the same across all OS's)

**Issue Descriptions:**

Friends often share shortlinks to GPM content with me, but there's no way to go to them from GPMDP, as there's a regex on the url that obviously won't match. In a perfect world, the url would be followed to make sure it's not a 301, and whatever the resolved url is would be matched against the regex.

Here's the offending line:
https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/d85ede4c771fb95bc9ae9e8bdf120953d5aac265/src/renderer/ui/components/modals/GoToModal.js#L58

**Steps to Reproduce:**

Here's a shortlink to some GPM content: https://goo.gl/T5j49Q

You can't follow that in GPMDP using ctrl+shift+g.

**Notes:**

I'd be willing to write up a PR for this, but wanted to see if it would be accepted first. I understand that this is sorta niche, but it would be nice for me and my friends.

rattboi:
Fixes #2015 

I figure this is still WIP because I threw the resolver function within gotomodal,js, but I honestly don't know where it's supposed to go. JS isn't really my language, so if I could get some direction, that'd be great. 

Also, I used the already-installed library "request", and wrapped it in a promise. I might be able to remove some of the code if I used the "request-promise" library instead, but I was trying to be as lightweight as possible. Again, any direction would be useful.


src/renderer/ui/components/modals/GoToModal.js
```diff
@@ -4,6 +4,33 @@ import { findDOMNode } from 'react-dom';
...
+
+const urlresolve = function urlresolve(url) {
+  return new Promise((resolve, reject) => {
+    const options = {
+      url,
+      headers: {
+        followAllRedirects: true,
+      },
+    };
+    request
```
MarshallOfSound: This sends requests from the `http` stack of node.js and some GPM URL's might require authentication to fully resolve (or they might redirect to the sign in page).  We should use `fetch` or the Electron `net` API.

src/renderer/ui/components/modals/GoToModal.js
```diff
@@ -54,10 +81,18 @@ export default class GoToModal extends Component {
       remote.app.quit();
     } else if (url === 'DEBUG_INFO') {
       Emitter.fire('generateDebugInfo');
+      this.handleClose();
+    } else {
```
MarshallOfSound: We should maintain the pass-quick logic from before (If it matches the RegExp use the URL straight away otherwise run the resolver).

src/renderer/ui/components/modals/GoToModal.js
```diff
@@ -4,6 +4,33 @@ import { findDOMNode } from 'react-dom';
 import Dialog from './ThemedDialog';
 import FlatButton from 'material-ui/FlatButton';
 import TextField from 'material-ui/TextField';
+import request from 'request';
+
+// Initialize the global Logger to forward to the main process.
+window.Logger = remote.getGlobal('Logger');
```
MarshallOfSound: The Logger should already be pulled in to the `renderer` process, I'd like to limit `remote` calls.  Also in general don't inject something into a global that is only being used in this file.

src/renderer/ui/components/modals/GoToModal.js
```diff
@@ -4,6 +4,33 @@ import { findDOMNode } from 'react-dom';
...
+
+const urlresolve = function urlresolve(url) {
```
MarshallOfSound: This should be a class method on `GoToModal` so that it is scoped correctly.

src/renderer/ui/components/modals/GoToModal.js
rattboi: Should we then verify it again, after resolution? I assume yes, so I should refactor out the test into its own function.

MarshallOfSound: Yes and Yes 👍

rattboi:
Ok, so I made all the easy changes that you requested. I'm hoping for a new review to make sure I'm headed in the right direction. I realize there will need to be new test cases, since I split out more methods, and I'll figure that out (haven't done tests in JS). 

I'm unclear on the net/fetch change. It looks like fetch in the codebase is node-fetch, so isn't that still in the node context and not the electron context? Will it be authed appropriately if I use node-fetch?

MarshallOfSound:
@rattboi I'll take a quick look shortly but RE `node-fetch`.

You don't need to use `node-fetch` in the renderer process, `fetch` will already be globally available and that *should* use the authentication from GPM.  You might need to use the `electron.net` module but hopefully not.

test/electron-renderer/ui/modals/GoToModal_spec.js
```diff
@@ -1,13 +1,18 @@
...
+import chaiAsPromised from 'chai-as-promised';
+
 import sinon from 'sinon';
 import { given } from 'mocha-testdata';
 
 import GoToModal from '../../../../build/renderer/ui/components/modals/GoToModal';
 
 import createModalTest from './_createModalTest';
 
+window.Promise = Promise;
```
MarshallOfSound: This is not needed

test/electron-renderer/ui/modals/GoToModal_spec.js
```diff
@@ -74,6 +79,77 @@ createModalTest(GoToModal, ['gotourl'], [], [], () => {}, (_c) => {
...
+      window.fetch.onSecondCall().returns(fakeRedirect());
+      window.fetch.onThirdCall().returns(fakeOk());
+    });
+
+    afterEach(() => {
+      window.fetch.restore();
+    });
+
+    given(['https://goo.gl/UFT1HU', 'https://goo.gl/ySV9WV'])
+    .it('should emit the navigate event if it is a valid redirected GPM URL', (url) => {
```
MarshallOfSound: This indenting looks weird, probably better to put it all on one line and if required ignore the eslint rule on this line

src/renderer/ui/components/modals/GoToModal.js
```diff
@@ -41,6 +41,47 @@ export default class GoToModal extends Component {
     }
   }
 
+  createRequest = (urlToResolve) =>
+    new Request(urlToResolve, {
```
MarshallOfSound: Today I learned, cool trick 👍

rattboi:
Woot, full coverage. Also, it looks like Coveralls isn't configured correctly? Is that why it's saying "Changes Unknown"?

test/electron-renderer/ui/modals/GoToModal_spec.js
```diff
@@ -74,6 +77,105 @@ createModalTest(GoToModal, ['gotourl'], [], [], () => {}, (_c) => {
     fired.generateDebugInfo.should.be.ok;
   });
 
+  describe('stubbed fetch request', () => {
```
MarshallOfSound: This is "context" to the nested `it` test.  Concatenate the strings to get what the logs will output.  This would make more sense as

```js
describe('with a redirecting URL', () => {`
```

Then in the logs you get

`GoToModal with a redirecting URL given URL should emit the ....`

PR #2059, opened 2016-12-30
Improved UI/UX of general settings tab

levrik:
Light theme: 
image

Dark theme:
image

### DIFF
--- a/src/renderer/ui/components/settings/ToggleableOption.js
+++ b/src/renderer/ui/components/settings/ToggleableOption.js
@@ -16,10 +16,16 @@ class SettingsCheckbox extends Component {
   }
 
   render() {
+    const label = (
+      <span>
+        {this.props.label}
+        {this.props.hintLabel ? <span className="settings-toggle-hint-label">{this.props.hintLabel}</span> : null}
+      </span>
+    );
+
     return (
       <Checkbox
-        label={[this.props.label, this.props.hintLabel
-          ? <span id="checkbox-hint" style={{ display: 'block', fontSize: '0.75em', color: 'gray' }}>{this.props.hintLabel}</span> : null]}
+        label={label}
         checked={this.props[this.props.settingsKey]}
         onCheck={this.onChange}
         style={{
--- a/src/renderer/ui/components/settings/ToggleableOption.js
+++ b/src/renderer/ui/components/settings/ToggleableOption.js
@@ -18,7 +18,8 @@ class SettingsCheckbox extends Component {
   render() {
     return (
       <Checkbox
-        label={[this.props.label, this.props.hintLabel ? <span style={{ display: 'block', fontSize: '0.75em', color: 'gray' }}>{this.props.hintLabel}</span> : null]}
+        label={[this.props.label, this.props.hintLabel
+          ? <span id="checkbox-hint" style={{ display: 'block', fontSize: '0.75em', color: 'gray' }}>{this.props.hintLabel}</span> : null]}
         checked={this.props[this.props.settingsKey]}
         onCheck={this.onChange}
         style={{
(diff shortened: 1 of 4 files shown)


src/renderer/ui/components/settings/ToggleableOption.js
```diff
@@ -17,7 +18,8 @@ class SettingsCheckbox extends Component {
   render() {
     return (
       <Checkbox
-        label={this.props.label}
+        label={[this.props.label, this.props.hintLabel
```
MarshallOfSound: This will throw warnings in the console (if it doesn't it should) due to not setting a key on each item in an inline array.

I'd also rather that this ternary expression was pulled out of the return call to a variable assignment

src/renderer/ui/components/settings/ToggleableOption.js
```diff
@@ -17,7 +18,8 @@ class SettingsCheckbox extends Component {
...
+          ? <span id="checkbox-hint" style={{ display: 'block', fontSize: '0.75em', color: 'gray' }}>{this.props.hintLabel}</span> : null]}
```
MarshallOfSound: Why does this span have an ID?  ID's are by definition supposed to be unique when we could have multiple of these on the page.

It would make more sense for this to be a class and then to apply the styling through the LESS stylesheets

levrik: Added the ID for targeting in the test. But I can change it to a class, style with that and use that for targeting in the test :)

src/renderer/ui/components/settings/ToggleableOption.js
levrik: Fixed.

levrik:
@MarshallOfSound Swapped screenshot with a newer one :)

levrik:
@MarshallOfSound Please wait before merge. Need to fix some issues with theming!

levrik:
@MarshallOfSound Okay. Theming support is done (see above). Ready to merge :)

src/renderer/ui/components/settings/ToggleableOption.js
```diff
@@ -9,6 +9,11 @@ class SettingsCheckbox extends Component {
     hintLabel: PropTypes.string,
     settingsKey: PropTypes.string.isRequired,
     setSetting: PropTypes.func.isRequired,
+    muiTheme: PropTypes.object,
+  };
+
+  static contextTypes = {
```
MarshallOfSound: Should really use the helper from MUI

https://github.com/callemall/material-ui/blob/master/src/styles/muiThemeable.js

levrik: Ah. No problem. It's been some time since I used MUI the last time.

levrik: @MarshallOfSound Okay. Done.

src/renderer/ui/components/settings/ToggleableOption.js
```diff
@@ -1,23 +1,35 @@
 import React, { Component, PropTypes } from 'react';
+import muiThemeable from 'material-ui/styles/muiThemeable';
 import Checkbox from 'material-ui/Checkbox';
 
 import SettingsProvider from '../generic/SettingsProvider';
 
-class SettingsCheckbox extends Component {
+const SettingsCheckbox = muiThemeable()(class extends Component {
```
MarshallOfSound: cc / the linter doesn't like this.

It would look better as

```js
class SettingsCheckbox extends Component {
  ..
}

const ThemedSettingsCheckbox = muiThemeable()(SettingsCheckbox);
```

And then just use `ThemedSettingsCheckbox`

levrik: Ah damn. Changed :)

PR #2138, opened 2017-01-20
Updated German translation

levrik:
I translated it with "Show current song on the lock screen". I think many people have no idea what the Windows 10 System Media Service is.
If you want the German translation to match the English sentence, please let me know and I will change it :)

### DIFF
--- a/src/renderer/ui/components/generic/PlatformSpecific.js
+++ b/src/renderer/ui/components/generic/PlatformSpecific.js
@@ -1,14 +1,20 @@
 import { Component, PropTypes } from 'react';
+import os from 'os';
 
 export default class PlatformSpecific extends Component {
   static propTypes = {
     children: PropTypes.object,
     platform: PropTypes.string.isRequired,
+    version: PropTypes.string,
   };
 
   render() {
     if (process.platform === this.props.platform) {
-      return this.props.children;
+      if (!this.props.version) return this.props.children;
+
+      if (os.release().startsWith(this.props.version)) {
+        return this.props.children;
+      }
     }
     return null;
   }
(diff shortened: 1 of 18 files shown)


MarshallOfSound:
> I translated it with "Show current song on the lock screen". I think many people have no idea what the Windows 10 System Media Service is.

The original reason I didn't use a string like that for the english version is that the behavior could change at any point without me doing anything.  I.e. the track appearing on the lock screen is entirely dependent on Windows 10.  Future versions may eliminate the behavior or change it.

A good compromise would be to leave the label what is already is but add a hint label with content similar to "This will show the current track on the lock screen".

levrik:
@MarshallOfSound Okay. Changed it :)

src/renderer/ui/components/generic/PlatformSpecific.js
```diff
@@ -1,14 +1,20 @@
...
   static propTypes = {
     children: PropTypes.object,
     platform: PropTypes.string.isRequired,
+    version: PropTypes.string,
   };
 
   render() {
     if (process.platform === this.props.platform) {
-      return this.props.children;
+      if (!this.props.version) return this.props.children;
```
MarshallOfSound: If we're making a version check part of the components API it should probably be a semver check.  I'll fix it for now because I'm bored 😆 but in the future can we avoid scope creep on PR's.  I.e. A translation change shouldn't affect the component API.

levrik: @MarshallOfSound Yeah. I implemented it first using `semver` but it seemed a bit overkill to me.

MarshallOfSound: @levrik I'm OK with overkill as long as it doesn't add too much file size to the overall package.  `semver` is like a few kb but it adds a lot more customizability.  I have an eventual plan to publish some of the internal React components of GPMDP as their own separate "Electron React Helpers" kind of package and using a standard like semver would be very good for if / when that happens

Issue #2085, opened 2017-01-07
Add option for removing tray icon


wa4557:
**OS:**
Ubuntu 16.04

**Issue Descriptions:**
It would be great if there was an option to hide the tray icon. On Ubuntu (and other linux desktop environemtns) you have SoundMenu integration so having both is a bit redunandant (tray and soundmenu). Could you add an option to remove the tray icon completely, which means that the app is no longer closing to the tray but rather closing to the soundmenu

**Steps to Reproduce:**
Open the app :)

jostrander:
Unfortunately this just isn't in the cards at the moment. 

See #1563 #1574 #1269 #1266

Issue #1574, opened 2016-09-02
Hide Menu Bar Icon on OSX


MartSerra:
**OS:**
OS X 10.11.6 El Capitan

**Issue Descriptions:**
Is it possible to hide the icon en Menu Bar? But not closing the application.

**Steps to Reproduce:**
An option for hide menu bar icon from _GooglePlayMusicDesktopPlayer_ without closing the application.

MarshallOfSound:
When you say Menu Bar.  Can you clarify exactly what you mean?

Do you mean the Dock or the System Tray?

MartSerra:
O my apologize, I mean hide the icon on system tray.
![screen shot 2016-09-03 at 12 03 13 pm](https://cloud.githubusercontent.com/assets/7231751/18226404/8745529a-71ce-11e6-9079-4c9554bcfc58.png)
And I would be awesome if the controls that are implemented on this menu tray icon could be placed on application dock.
![screen shot 2016-09-03 at 12 09 19 pm](https://cloud.githubusercontent.com/assets/7231751/18226435/6d6d5d1c-71cf-11e6-8a3d-53da722c7cb8.png)
 For example I attached a Spotify controls.
![screen shot 2016-09-03 at 12 05 30 pm](https://cloud.githubusercontent.com/assets/7231751/18226417/d62a6404-71ce-11e6-8713-4f6eab90a52c.png)
Thanks for your support @MarshallOfSound

MarshallOfSound:
@MartSerra I've had the "please let me hide the system tray icon" discussion more times than I can count on the issues section on GitHub.

TLDR: We can't and we won't 😢

Issue #1269, opened 2016-07-04
Show/hide appliaction to tray icon


francma:
**OS:**
Linux with i3wm (tilling WM) and i3bar

**Application Version:**
3.3.0

**Issue Descriptions:**
Show/hide option from tray icon menu doesn't work. It always says "show" despite fact it's already shown. Double clicking tray icon and closing application does nothing.

I tested it with similar electron app (twolfson/google-music-electron) and it worked (so it's not WM or Electron issue) -- clicking icon toggles shown/hidden (same with tray icon menu option). Maybe option "Close to Tray" would do the trick (since there is no minimize in tilling WM).

Thanks :)

MarshallOfSound:
> Show/hide option from tray icon menu doesn't work. It always says "show" despite fact it's already shown. 

This is intended behaviour and in fact no one has ever considered this an issue before.  The button "shows" the player, whether the player is shown or not is not releveant.  I don't consider that a bug?

> Double clicking tray icon and closing application does nothing. 

Can you clarify what you mean here.  There was a massive thread a while ago where a conclusion was reached that on linux a double click should be a noop and a single click should toggle the window.

> I tested it with similar electron app (twolfson/google-music-electron) and it worked (so it's not WM or Electron issue)

Everything you have described comes down to individual implementation, all the buttons and click actions we handle manually and it is down to the application to handle them and respond as it sees fit.  Just because one electron app does something (twolfsen) doesn't mean another will 😄 

> Maybe option "Close to Tray" would do the trick (since there is no minimize in tilling WM).

This exists, have you tried clicking the close button on the window 😆   By default it will minimize to tray unless you turn that feature off in settings

francma:
Single clicking doesn't work too. Closing application does nothing like I already said.

The thing is there is no minimize or close button in tilling WMs (it sends close via command). However, I haven't had any problems with other apps yet.

MarshallOfSound:
> Closing application does nothing like I already said.

When you close the window through a command do you know if it is sending a close window command or a quit application command?

If it sends a close window command then it would minimize to system tray, if it sends a quit command it would quit the entire application.

This application was never designed with tiling window managers in mind and I personally have never used one so I'm not familiar with usage or expected actions.

francma:
This issue is similar to https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/issues/270.

[...]

Issue #1266, opened 2016-07-04
Request: Add option to hide system tray icon


YuukiMinoru:
There is an option to minimise to system tray, but no option to not use the system tray at all.

MarshallOfSound:
This has been raised a few times and declined every time.  Lots of functionality depends on the tray icon and it plays a critical role in controlling the app

YuukiMinoru:
Perhaps this should be added to the FAQ? If its coming up frequently.

PR #2143, opened 2017-01-21
Feature/hide tray icon

iDontCYnot:


MarshallOfSound:
https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/master/CONTRIBUTING.md

> Please ask first before embarking on any significant pull request (e.g. implementing features, refactoring code, porting to a different language), otherwise you risk spending a lot of time working on something that the project's developers might not want to merge into the project.

I have repeatedly rejected this idea

#2085 #1574 #1269 #1266 #636 #300

And I stand by my decision in all of those issues / PR's, this does not account for all scenarios and this in its current state leave a user with an un-quitable or un-openable application.

E.g. Window is "minimized to tray" and the user opens the desktop settings window.  Disables the tray icon, the tray icon dissapears.  Now how do they get their window back?

It's small edge cases like that ^^ and the fact the tray icon is fundamentally tied to window management / application quitting on 2 out of 3 platforms that adding an option to disable it will add more trouble than it is work.

While I appreciate PR's from the community this idea has been struck down time and time again due to it being inherently complicated to change this behavior.

jostrander:
With @MarshallOfSound's disapproval I'm going to close this.

PR #2147, opened 2017-01-22
Update italian translation

voria:
This change is [Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/2147)


voria:
Update PR to include latest added strings.

PR #2162, opened 2017-01-25
Update pl-PL.json

Kun2210:
Updated several missing translations


Kun2210:
Honestly, I don't understand the reason for the check to fail. Any hint?

MarshallOfSound:
@Kun2210 The issue is present in master as well 👍

PR #2176, opened 2017-01-28
add Japanese language

jsoizo:
This change is [Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/2176)


src/renderer/ui/utils/constants.js
```diff
@@ -48,6 +48,7 @@ export const languageMap = {
   'en-US': 'English',
   'fr-FR': 'French',
   it: 'Italian',
+  ja: '日本語',
```
MarshallOfSound: Historically we have the English name of the Language here, so this should be `Japanese`.

Though in the future I guess these strings should be translatable as well 🌵 

But for now (and for consistency) can we set this to be Japanese 😄

jsoizo: Sorry, I fixed it.

Issue #2103, opened 2017-01-14
Version 4.1.1 doesn't run, gives white screen on Ubuntu


philh7456:
I previously had 4.0.5 running fine on Ubuntu 16:10, 64bit. Upgraded to 4.1.1 and GPMDP either does not start or starts but just gives a white screen with no content. I have tried deleting the config files and starting again, but no luck.
Starting from the command line shows the following:

Starting Sentry
error: Failed to load bonjour with error: {}
Bonjour is required to use Chromecast Support or to enable ZeroConf for the PlaybackAPI
On linux you need to install "avahi"
Failed to encode dictionary
Failed to encode reply value
Failed to encode dictionary
Failed to encode reply value

avahi-daemon is installed and up-to-date (0.6.32-1ubuntu1).

I have tried generating a debug file, but pressing Ctrl-shift-g in the white window does nothing.

Downgraded to 4.05, works fine.

MarshallOfSound:
Looks like a transitive dependency bump broke something.  I would pull the linux update but I can't as it isn't technically an update (people download manually).  I've pulled the bintray release though.

/cc @jostrander can you investigate?

oleerik:
Same issue on fedora25 :cactus:

MarshallOfSound:
I deleted the uploaded `4.1.1` linux artifacts so no one else will download into this issue.

voria:
I've just built the project from the current master branch (c3081d2) on Fedora 25 x86_64, starting from a clean tree. The resulting rpm from the command `npm install && npm run make:rpm:64` works perfectly fine, no issues whatsoever.

MarshallOfSound:
@voria To confirm, building locally for you works as expected but the previously published builds did not?

voria:
@MarshallOfSound I had no chance to try the published builds, they were already taken off when I noticed the new 4.1.1 release. If they are still available for download somewhere, I can test them.

MarshallOfSound:
Try this one out for size 👍 

https://1489-40008106-gh.circle-artifacts.com/3/home/ubuntu/Google-Play-Music-Desktop-Player-UNOFFICIAL-/dist/installers/redhat/google-play-music-desktop-player-4.1.1.x86_64.rpm

voria:
It does NOT work, exactly as reported. At start, GPMDP gives a white screen with no content. This is the terminal output (nothing relevant I think):

> Starting Sentry
[...]

PR #2186, opened 2017-01-31
Ignore linux for version range checks.

BrianAllred:
Patch fixes issue #2103.

### DIFF
--- a/src/renderer/ui/components/generic/PlatformSpecific.js
+++ b/src/renderer/ui/components/generic/PlatformSpecific.js
@@ -3,7 +3,7 @@ import os from 'os';
 import semver from 'semver';
 
 const parsedOSVersion = semver.parse(os.release());
-const osVersion = `${parsedOSVersion.major}.${parsedOSVersion.minor}.${parsedOSVersion.patch}`;
+const osVersion = parsedOSVersion ? `${parsedOSVersion.major}.${parsedOSVersion.minor}.${parsedOSVersion.patch}` : null;
 
 export const semverValidator = (props, propName, componentName) => {
   if (props[propName]) {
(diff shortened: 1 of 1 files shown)


BrianAllred:
Ha, that was how I originally had it and second guessed myself.

jostrander:
I'd defer to @MarshallOfSound before you change it, he might not mind. Just quickly glancing over it, it seems to be an appendage without explanation.

src/renderer/ui/components/generic/PlatformSpecific.js
```diff
@@ -23,7 +23,7 @@ export default class PlatformSpecific extends Component {
     if (process.platform === this.props.platform) {
       if (!this.props.versionRange) return this.props.children;
 
-      if (semver.validRange(this.props.versionRange) && semver.satisfies(osVersion, this.props.versionRange)) {
+      if ((semver.validRange(this.props.versionRange) && semver.satisfies(osVersion, this.props.versionRange)) || (osVersion === null && process.platform === 'linux')) {
```
MarshallOfSound: If `osVersion` is `null` we should not render regardless of platform.  Also this appears to short-circuit the wrong way.  The `semver.satisfies` check will called on `null` still.

Reverse the "or" order and don't check the platform 👍 

Trying to keep this components in a state where they could be published separately

voria:
This does not fix the problem. As I described in the issue, the problem is mainly at lines 5-6 of the file `src/renderer/ui/components/generic/PlatformSpecific.js`, where `parsedOSVersion` is `null` and then there is an attempt to read its properties, causing an uncaught error.

BrianAllred:
@voria Interesting, I had the white screen issue and this fixed it for me. Maybe when I make @MarshallOfSound's suggested changes, it'll fix it for you.

SriRamanujam:
fwiw this patch does fix #2103 for me on git master, thank you!

Issue #921, opened 2016-04-08
Last.FM doesn't work with proxy settings


rattboi:
**OS:**
Windows 7

**Issue Descriptions:**
Last.fm is blocked, so I have a socks proxy I use. I only want the traffic for GPMDP going through the proxy. I've enabled my proxy by adding '--proxy-server="socks5://localhost:8089"' to the gpmdp shortcut. I've tested that the music is using the proxy by stopping the proxy server, and seeing that music streaming no longer works, then starting the proxy server and streaming starts working again.

**Steps to Reproduce:**
1. Enable proxy (I know this isn't trivial for everyone) via --proxy-server="socks5://<proxyinfo>"
1. Go to "Desktop Settings -> Last.FM"
2. Click "Log in to Last.fm"
3. It shows "Authorization in progress" very briefly, then flips back to "Not Authorized"
4. It doesn't scrobble (obviously)

rattboi:
Perhaps it's related to this? https://github.com/electron/electron/issues/183

MarshallOfSound:
It is exactly that problem @rattboi and unfortunately that is quite tricky to fix.  It would require the app using handling proxy settings internally which is quite complex

jostrander:
Can we use something like https://www.npmjs.com/package/global-tunnel along with reading the command line settings to set the proxy for the node side of the web stack?

eythian:
I suspect close to this is:

    error: ################### !! Update server down !! ##################

appearing on the terminal.

MarshallOfSound:
I'm removing this from the `v4` milestone and will implement using  the new `net` module that is coming to Electron --> https://github.com/electron/electron/pull/7577

rattboi:
FYI, looks like that PR got merged into Electron 15 days ago

MarshallOfSound:
@rattboi I am aware :P

asermax:
Any news on this? 
Is there any workaround to generate the credentials that we could use to have it working? I don't mind meddling with some code if necesary.

MarshallOfSound:
@asermax It's not just the credentials that are the issue, the actual REST calls to trigger last.fm actions will face the same proxy issue.  This isn't on my immediate todo list but if someone wants to pick it up that would be great 😄

asermax:
Ooo, it makes sense :thinking: 
[...]

PR #2241, opened 2017-02-17
Use electron.net on lastfm requests

asermax:
Fixes related to #921
This change monkey-patches the external `lastfm-node` library to use `electron.net` instead of node's `http` module.

### DIFF
--- a/src/main/features/core/lastFM.js
+++ b/src/main/features/core/lastFM.js
@@ -6,7 +6,7 @@ import { LASTFM_API_KEY, LASTFM_API_SECRET } from '../../constants';
 // hijack lastfm module to use 'electron.net' instead of 'http'
 global.GENTLY_HIJACK = {
   hijack(require) {
-    return function (moduleName) {
+    return (moduleName) => {
       let module;
 
       if (moduleName === 'http') {
--- a/src/main/features/core/lastFM.js
+++ b/src/main/features/core/lastFM.js
@@ -1,8 +1,29 @@
 import { BrowserWindow } from 'electron';
-import { LastFmNode } from 'lastfm';
 import path from 'path';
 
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
@asermax Can you confirm if this works or not? 👍

src/main/features/core/lastFM.js
```diff
@@ -1,9 +1,30 @@
 import { BrowserWindow } from 'electron';
-import { LastFmNode } from 'lastfm';
 import path from 'path';
 
 import { LASTFM_API_KEY, LASTFM_API_SECRET } from '../../constants';
 
+// hijack lastfm module to use 'electron.net' instead of 'http'
+global.GENTLY_HIJACK = {
+  hijack(require) {
+    return function (moduleName) {
```
MarshallOfSound: `function` requires a name 👍

asermax:
@MarshallOfSound I did a test it earlier today but with no success. Not being able to debug is a real pain too :( I didn't clean the previous installation when testing it tho, so I'll completely remove the player from my comp and give it another try tomorrow just in case and get back to you.
Beyond the eslint issue, does this change look ok? Can you think of any reason why it could be failing?

src/main/features/core/lastFM.js
asermax: I'll change it for an arrow function since it seems that's the norm on the code (I didn't realize earlier 🙇)

asermax:
Tried again with no success :(

rattboi:
@asermax Any update on this issue? Just started a new job with http proxy and running into this again, lol.

asermax:
@rattboi not really, I did some tests with no success and came back to using the web app with a chrome plugin for lastfm :<

MarshallOfSound:
Closing due to lack of activity, if you still wish this PR to go through please rebase onto current master and follow up with the provided comments

PR #2244, opened 2017-02-19
Add info track keyboard shortcut

wgrafael:
Simple feature but useful that add a keyboard shortcut that when fired sends a notification with information of current track.
(with portuguese translation)

captura de tela de 2017-02-18 23-01-45
captura de tela de 2017-02-18 23-41-54

### DIFF
--- a/src/_locales/en-US.json
+++ b/src/_locales/en-US.json
@@ -62,6 +62,7 @@
   "playback-label-thumbs-up": "Thumbs Up",
   "playback-label-volume-down": "Decrease Volume",
   "playback-label-volume-up": "Increase Volume",
+  "playback-label-info-track": "Notify current track",
   "playback-os-no-track-playing": "No Track Playing",
 
   "settings-requires-restart": "Requires restart",
--- a/src/renderer/windows/GPMWebView/playback/controller.js
+++ b/src/renderer/windows/GPMWebView/playback/controller.js
@@ -78,9 +78,11 @@ Emitter.on('playback:miniDisable', () => {
 });
 
 Emitter.on('playback:infoTrack', () => {
-  if (!remote.getGlobal('PlaybackAPI').data.song.title) return;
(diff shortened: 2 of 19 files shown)


src/renderer/windows/GPMWebView/playback/controller.js
```diff
@@ -76,3 +76,11 @@ Emitter.on('playback:miniEnable', () => {
 Emitter.on('playback:miniDisable', () => {
   window.GPM.mini.disable();
 });
+
+Emitter.on('playback:infoTrack', () => {
+  if (!remote.getGlobal('PlaybackAPI').data.song.title) return;
```
MarshallOfSound: This will throw errors as `song` can be null

src/renderer/windows/GPMWebView/playback/controller.js
```diff
@@ -76,3 +76,11 @@ Emitter.on('playback:miniEnable', () => {
...
+  new Notification(remote.getGlobal('PlaybackAPI').data.song.title, { // eslint-disable-line
```
MarshallOfSound: These remote calls are taxxing on the UI and not required.  The information we need is available in this process through the `GPM` global.

https://github.com/gmusic-utils/gmusic.js#playbackgetcurrenttrack

```js
GPM.getCurrentTrack();
``

**Comment by wgrafael** [comment]
Thank you for your review. I committed the change.

File: src/_locales/en-US.json [hunk 3]
```diff
@@ -62,6 +62,7 @@
   "playback-label-thumbs-up": "Thumbs Up",
   "playback-label-volume-down": "Decrease Volume",
   "playback-label-volume-up": "Increase Volume",
+  "playback-label-info-track": "Notify current track",
```
MarshallOfSound: The tests will fail without an empty translation string in all the other language files.

MarshallOfSound:
@wgrafael One last thing to fix the tests I think 👍

PR #2268, opened 2017-02-27
Add Snapcraft build metadata

RAOF:
Snaps (https://snapcraft.io/) are a Linux application distribution format.

This adds the requisite metadata to build a GPMDP snap from the repository.
The major features work - MPRIS integration, desktop indicators/systray, music hotkeys and so on.

The desktop file icon doesn't work, and you will get an all-white window unless you manually connect the “process-control” snap interface.

With this metadata in the repository we can get Launchpad (http://launchpad.net/) to do daily builds of the GPMDP snap and publish them to “edge” channel of the snap store.


snap/gui/google-play-music-desktop-player.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
```
MarshallOfSound: Can this be moved to `vendor` and somehow indicate that it is related to `snap`.

Something like `vendor/snap/google-play-music-desktop-player.desktop`

.gitignore
```diff
@@ -9,5 +9,13 @@ vendor/*.msi
 electron.xml
 spectron.xml
 coverage.zip
-uwp/
```
MarshallOfSound: This still needs to be here

snapcraft.yaml
```diff
@@ -0,0 +1,58 @@
...
+description: |
+  This Google Play Music player provides desktop integration - with music
+  hotkeys and the sound indicator via an MPRIS interface, the abiliy to use
+  custom colours and load custom themes, a remote-control interface, and
+  lower resource consumption than a full-blown browser tab.
+  .
+  This is *not* an official Google project.
+grade: stable
+confinement: strict
+icon: snap/gui/google-play-music-desktop-player.png
```
MarshallOfSound: Can this point to the PNG generated as part of the build phase?

RAOF: Sadly, no. The `snapcraft` tool checks for the presence of the `icon:` file at the beginning of the build.

Oh. But it accepts SVG, so it *could* point to `src/assets/icons/svg/vector_logo.svg`. Is there any reason not to do that?

RAOF:
To get this building and deploying I can add a Launchpad import of the github repository; that will pull the branch four times a day and build + publish the snap (to the “edge” channel) if there are any changes.

If you'd prefer to do a build per commit, the `snapcraft enable-ci travis` command is supposed to be able to do the necessary changes to add a snapcraft build to Travis.

Launchpad has the advantage that it won't slow down CI and we'll be able to extend this (after a couple of snapcraft bugs have been fixed) to build packages for arm, arm64, i386, and (for the truly weird) PowerPC. It's got the disadvantage that changes which break the snap build won't be noticed by CI.

Once the builds are in the edge channel, you can promote them to the beta/candidate/stable channels from the developer page on the snap store (or, at least, you'll be able to once I've set that up 😏). There are other possibilities, like setting the git-tag to build from in the `snapcraft.yaml` file, or manually building locally and uploading, if you'd prefer.

MarshallOfSound:
If you delete the PNG and rebase we can try get this going for `4.2.0`

Ads20000:
Ah sorry I filed an issue before checking closed issues/PRs, silly me, @RAOF how's setting up the developer page on the Snap store for @MarshallOfSound to use? They're adding a feature which will allow you to add additional contributors to a Snap which I think might help, but as far as I know that's still in the works...

PR #2275, opened 2017-03-02
Fix misspelling in postinstall script definition

jaycegregg0:
optional was spelled 'optioanl'


MarshallOfSound:
Well this is just embarrassing 😢

PR #2293, opened 2017-03-06
Add needed apostrophe

curtisgibby:
"computer's resources" needs an apostrophe


MarshallOfSound:
We don't actually use the `gh-pages` branch, rather the `website` branch

https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/tree/website

I can make the change if you don't want to do it again for such a small thing 👍

PR #2337, opened 2017-03-20
replacing "not set" in hotkey-settings with a locale string

ChrisB9:
+ adding a missing string in de locale


ChrisB9:
i'll add the corrected test soon

test/electron-renderer/ui/HotkeyInput_spec.js
```diff
@@ -39,7 +39,7 @@ describe('<HotkeyInput />', () => {
 
   it('should render a text field with the correct default value', () => {
     const component = mount(<HotkeyInput label="Test Label" hotkeyAction="playPause" />, materialUIContext);
-    component.find(TextField).props().value.should.be.equal('Not Set');
+    component.find(TextField).props().value.should.be.equal(undefined);
```
MarshallOfSound: This is what is breaking the tests currently, it should be expecting "settings-option-hotkey-not-set"

PR #2355, opened 2017-03-22
Fix back swiping

stevenla:
On Mac, 3-finger swiping backwards didn't work, but swiping forwards did. This PR fixes that issue.


stevenla:
Is the travis test suite known to be flakey? I don't think the test failure has anything to do with my commit

MarshallOfSound:
Hm, will let the codeclimate check slide this time.  Seems to be a weird interpretation of duplicate code 👍

PR #2562, opened 2017-06-01
Add F keys up to F24

### LINKED ISSUE #2556: Support nonstandard keys
**OS:** Windows 10 1703 (build 15063.296)

**Issue Descriptions:** Unusual keys such as F13, F14 and F15 are unsupported in GPMDP. I'm generating these keycodes with an arduino based macro keypad and would like to bind them as like/dislike without interfering with any other things on the machine.

**Steps to Reproduce:** Produce F13/F14/F15 scancodes and try to bind them as things in the settings.

Note: I've tested these with [this thing](http://www.delphiforfun.org/programs/utilities/KeyCodes.htm) and they definitely are appearing.

lf-:
image

Now if only I didn't have to press CTRL..... I'll hack it on the arduino side, but it would be a lot better if I didn't need to do that.

Fixes #2556


MarshallOfSound:
Seems legit 👍

Issue #2545, opened 2017-05-24
Exit on Ctrl+Q shortcut


agurenko:
It's just a simple feature request to respond to Ctrl+Q shortcut and exit application. Since most of the apps under linux has ctrl+q as a quit/exit shortcut GPMDP is one of the very few apps that don't behave :)

PR #2563, opened 2017-06-01
Added shortcut for 'close', related to #2545

sk0g:
Added an extra label, 'Close2', to fix issue described in #2545
There was no Linux detection as of yet, so not sure if it is meant to be done this way.

### DIFF
--- a/src/main/features/core/applicationMenu.js
+++ b/src/main/features/core/applicationMenu.js
@@ -128,6 +128,11 @@ const template = [
         accelerator: 'CmdOrCtrl+W',
         role: 'close',
       },
+      {
+        label: 'Close2',
+        accelerator: 'CmdOrCtrl+Q',
+        role: 'close',
+      },
     ],
   },
   {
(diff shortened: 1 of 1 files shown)


src/main/features/core/applicationMenu.js
```diff
@@ -125,7 +125,10 @@ const template = [
       },
       {
         label: 'Close',
-        accelerator: 'CmdOrCtrl+W',
+        accelerator: (() => {
```
MarshallOfSound: This creates a new function scope needlessly, why not use the ternery I suggested in the previous comment?

sk0g: Ah sorry, I thought you meant that as a general guideline and noticed you ad already done something similar, so tried to keep it consistent. I'll do another one real quick.

MarshallOfSound: I apologize, following existing coding style is always a good thing, looks like I made some terrible code back in the day 😆

PR #2584, opened 2017-06-17
Add application menu keyboard shortcut for showing album cover

calmandniceperson:
This PR adds a keyboard shortcut (Cmd/Ctrl+Shift+C) to the application menu, allowing users to show the album cover / screensaver via their keyboard. I know this may seem like something unimportant, but I, for one, use the album cover function all the time and thought it would be a nice addition, considering the lyrics view can also be opened via a keyboard shortcut.

### DIFF
--- a/src/renderer/windows/GPMWebView/interface/customNavigation/showCover.js
+++ b/src/renderer/windows/GPMWebView/interface/customNavigation/showCover.js
@@ -1,0 +1,3 @@
+Emitter.on('cover:show', () => {
+  document.getElementById('hover-icon').click();
+});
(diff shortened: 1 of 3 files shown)


calmandniceperson:
Maybe renaming the menu point from "Show Album Cover" to "Toggle Screensaver" or "Toggle Album Cover" is better since it also closes it again? Wasn't sure so I just named it "Show Album Cover".

src/renderer/windows/GPMWebView/interface/customNavigation/showCover.js
```diff
@@ -0,0 +1,3 @@
+Emitter.on('cover:show', () => {
+  document.querySelector('#hover-icon').click();
```
MarshallOfSound: I would prefer that this is done in a safer way, google change selectors all the time. Just assigning the DOM node to a variable and checking the variable isn't null before clicking would be a better thing to do 👍

calmandniceperson:
@MarshallOfSound I changed it to check if the hover icon exists before trying to click. Unfortunately there doesn't seem to be another way of accessing the album view, as I'm sure you know. If any other changes come to your mind, let me know.

Issue #2590, opened 2017-06-21
Feature Request: Setting to hide track info/artwork from volume overlay


dotjosh:
**OS:** 
Windows 10

**Issue Description:**
The current "Windows 10 System Media Service" setting is great for getting media keys integrated with the app.  But along with it, the entire banner containing the track info is forced upon you and can take a considerable amount of screen real estate when you adjust the volume or play/pause a track.  

There is no global setting in Windows 10 to disable it.

![image](https://user-images.githubusercontent.com/180450/27366407-f2dabd68-5613-11e7-956c-97cee21a79e1.png)

**Feature Proposal:**
I think we should have an additional setting (_enabled by default to not change current default behavior_) for users who only wanted their media keys to work.

![image](https://user-images.githubusercontent.com/180450/27366574-029bc8d6-5615-11e7-9f5b-1ed55b359dc7.png)

dotjosh:
Implemented in pull request: #2591

PR #2591, opened 2017-06-21
Adds Windows 10 setting to hide track info/artwork in volume overlay

dotjosh:
This PR applies the feature described here:  #2590 

The new setting is only visible in Windows 10 and when the existing "enableWin10MediaService" setting is enabled.

### DIFF
--- a/src/renderer/ui/components/settings/ToggleableOption.js
+++ b/src/renderer/ui/components/settings/ToggleableOption.js
@@ -27,9 +27,9 @@ class SettingsCheckbox extends Component {
           : null}
       </span>
     );
-    const dependsOnSettingsKey = this.props.dependsOnSettingsKey;
+    const { dependsOnSettingsKey } = this.props;
     if (dependsOnSettingsKey && !this.props[dependsOnSettingsKey]) {
-      return <noscript />;
+      return null;
     }
 
     return (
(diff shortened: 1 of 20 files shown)


src/renderer/ui/components/settings/ToggleableOption.js
```diff
@@ -50,17 +55,22 @@ export default class ToggleableOption extends Component {
...
 
   render() {
     const checkboxProps = {
       label: this.props.label,
       hintLabel: this.props.hintLabel,
       settingsKey: this.props.settingsKey,
+      dependsOnSettingsKey: this.props.dependsOnSettingsKey,
     };
-
+    let keys = [this.props.settingsKey];
```
MarshallOfSound: Should be `const`

src/renderer/ui/components/settings/ToggleableOption.js
```diff
@@ -26,6 +27,10 @@ class SettingsCheckbox extends Component {
           : null}
       </span>
     );
+    const dependsOnSettingsKey = this.props.dependsOnSettingsKey;
```
MarshallOfSound: `const { dependsOnSettingsKey } = this.props;`

src/renderer/ui/components/settings/ToggleableOption.js
```diff
@@ -26,6 +27,10 @@ class SettingsCheckbox extends Component {
...
+    if (dependsOnSettingsKey && !this.props[dependsOnSettingsKey]) {
+      return <noscript />;
```
MarshallOfSound: You can return null from a component to render nothing, more optimal than mounting an unused DOM element.

`return null;`

MarshallOfSound:
@dotjosh Something weird's going on with these commits, looks like you somehow pulled in commits to this PR that are already on master.  I think a rebase is in order 😄

dotjosh:
Sorry about that - that should fix it.  By the way, this is impressive project management and development for someone your age.

PR #2611, opened 2017-06-27
Fix logo rendering on Gnome

TheBITLINK:
Some SVG engines (like librsvg on Gnome) don't seem to handle styles well.

For example, in Gnome 3.24, the logo gets rendered like this:

image
image

After replacing CSS styles with simple SVG attributes, the logo renders just fine

image
image


jostrander:
It really seems like this should be fixed in librsvg and not here. Styles are perfectly valid in SVG's. Maybe raise a bug with gnome's bugzilla? 

https://bugzilla.gnome.org/enter_bug.cgi?product=librsvg

MarshallOfSound:
SVG looks identical and if this fixes an issue I'll merge it for now, I would like it raised upstream though so other projects don't have to figure this out for themselves 👍

PR #2648, opened 2017-07-19
Feature/AddedSpanishLanguageSupport

markoshorro:
Added spanish language support (es-ES, spanish from Spain)


markoshorro:
How could I fix the Travis error?

src/renderer/ui/utils/constants.js
```diff
@@ -46,6 +46,7 @@ export const languageMap = {
   da: 'Danish',
   de: 'German',
   'en-US': 'English',
+  'es-ES': 'Spanish',
```
MarshallOfSound: There are trailing spaces on this line causing the tests to fail

markoshorro:
Why is it failing again?

PR #2684, opened 2017-08-03
Add the music media.role for PulseAudio

kabili207:
This sets the media.role label on media stream to "music" so PulseAudio's corking mechanism can work properly.

See this post on reddit for more detail on this. https://www.reddit.com/r/linux/comments/2rjiaa/horrible_decisions_flat_volumes_in_pulseaudio_a/cngt9a6/

### DIFF
--- a/vendor/snap/google-play-music-desktop-player.desktop
+++ b/vendor/snap/google-play-music-desktop-player.desktop
@@ -7,3 +7,4 @@ Icon=main.png
 Type=Application
 StartupNotify=true
 Categories=AudioVideo;Audio;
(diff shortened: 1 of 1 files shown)


SolarLiner:
Cannot reproduce the desired result here, under Ubuntu GNOME 17.04. I've added the X-PulseAudio-Properties line but sinks still don't get the proper media role.

Note that launching the app with PULSE_PROP=media.role=music as stated in the linked Reddit comment does add the role to the sink.

I believe a better way to add roles would be from whatever package wraps connection to PulseAudio, but that might be out of the scope of Electron itself.

PR #2688, opened 2017-08-05
Update slider when window is minimized

petuhovskiy:
This PR replaces requestAnimationFrame for always invoking function that updates slider.
Found such a function in GPM sources, 47th line checks for that function.

gmusic.js@6.1.1 works not so well for me and it seems using audio element for watching track progress is not enough.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "gmusic-mini-player.js": "^2.0.10",
     "gmusic-theme.js": "^2.1.15",
     "gmusic-ui.js": "^2.2.3",
-    "gmusic.js": "^6.1.1",
+    "gmusic.js": "^6.1.0",
     "html-entities": "^1.2.0",
     "jquery": "^2.2.4",
     "lastfm": "^0.9.2",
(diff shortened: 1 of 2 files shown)


src/renderer/windows/GPMWebView/index.js
```diff
@@ -37,6 +37,21 @@ const waitForExternal = setInterval(() => {
     require('gmusic-mini-player.js')(GMusic);
     const GMusicTheme = require('gmusic-theme.js');
 
+    // Replace requestAnimationFrame for invoking what we want, even when window is minimized
+    const defaultAnimationFrame = window.requestAnimationFrame;
+    window.requestAnimationFrame = (f) => {
+      if (f.toString) {
+        const str = f.toString();
```
MarshallOfSound: This is going to dramatically slow down animations, stringifying functions is a computation heavy task.

package.json
```diff
@@ -65,7 +65,7 @@
     "gmusic-mini-player.js": "^2.0.10",
     "gmusic-theme.js": "^2.1.15",
     "gmusic-ui.js": "^2.2.3",
-    "gmusic.js": "^6.1.1",
+    "gmusic.js": "^6.1.0",
```
wolfreak99: Why are you down grading the number?

petuhovskiy: gmusic.js@6.1.0 uses slider to get current playback position. Next gmusic.js versions use audio element instead (for now) and updating slider has no sense then

MarshallOfSound:
Fixed in master

Issue #2380, opened 2017-03-27
Speaker fill does not work with desktop app


sergesyrota:
**OS:** Windows 10

**Issue Descriptions:** Speaker fill on 5.1 system does not work, and only front L/R speakers are playing, while all others are silent.

**Steps to Reproduce:** Press Play. Disconnect front left/right speakers from amp. Ta da!

**Workaround:** Create a shortcut to "Google Play Music Desktop Player.exe" with "--try-supported-channel-layouts" flag. Ta da!

Unfortunately workaround does not work with the default shortcut to "Update.exe"

Would be nice to get this to work out of the box, or have this as a setting in the app itself.

MarshallOfSound:
@duplicate-issue-bot[bot] Heh, those are a bit off the mark 😆 

@sergesyrota I would accept a PR to add an option to the "Audio" settings tab to append this launch flag at startup.

`--try-supported-channel-layouts`

Would be a simple `Settings.get` call inside this file https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/master/src/main/configureApp.js 👍

sergesyrota:
@MarshallOfSound, I'll try this out, and if it works, will submit a PR.

sergesyrota:
After an hour trying (first time with nodejs), can't get the build to work. I'll get back to this some time in the future when I have more time, but if anyone else can try it, please do :)

MarshallOfSound:
@sergesyrota If you need help getting the build going feel free to ask in Gitter.

A good starting point is https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/wiki/Development-FAQ

MarshallOfSound:
Fixed in #2715

PR #2715, opened 2017-08-20
Add support for try-supported-channel-layouts flag

SkaceKamen:
This PR adds new setting value, which allow users to launch the player with `try-supported-channel-layouts` flag, which can fix 5.1 channel support for some people.

This change was discussed at #2380 and should fix it.

### DIFF
--- a/src/_locales/en-US.json
+++ b/src/_locales/en-US.json
@@ -114,6 +114,7 @@
   "settings-option-mini-use-scroll-volume": "Use the scroll wheel to adjust volume",
 
   "settings-option-skip-bad-songs": "Automatically skip disliked songs",
+  "settings-option-try-supported-channel-layouts": "Try supported channel layouts (requires restart, should fix 5.1 sound)",
 
   "settings-options-style-description": "Choose a local CSS file to inject styles into the application",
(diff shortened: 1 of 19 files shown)


src/_locales/en-US.json
```diff
@@ -114,6 +114,7 @@
   "settings-option-mini-use-scroll-volume": "Use the scroll wheel to adjust volume",
 
   "settings-option-skip-bad-songs": "Automatically skip disliked songs",
+  "settings-option-try-supported-channel-layouts": "Try supported channel layouts (requires restart, should fix 5.1 sound)",
```
MarshallOfSound: I think this needs to more non-tech friendly.  Maybe something like

"Experimental support for multiple audio channels such as 5.1 audio"

The "requires restart" bit should use the subLabel option and the "requires-restart" translation key that already exists

SkaceKamen: Makes sense, will fix

SkaceKamen: I can't really think anything better, will the phrase you proposed be sufficient?

MarshallOfSound: Yes

PR #2753, opened 2017-09-09
about-tab: Add a About tab to the Settings page.

jjerome00:
I can never figure out the version of the app installed on my computer.  I decided add it as a tab in the Desktop Settings section.  After getting it in there, it seemed simple enough that it might be worth sharing.

I'm rather new to React, but I was able to figure out the pattern from other parts of the code.  I'm also rather new to open source PR's so I apologize in advance if I missed something.

settings dialog

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,10633 @@
+{
+  "name": "google-play-music-desktop-player",
+  "version": "4.4.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "7zip": {
+      "version": "0.0.6",
+      "resolved": "https://registry.npmjs.org/7zip/-/7zip-0.0.6.tgz",
+      "integrity": "sha1-nK+xca+CMpSQNTtIFvAzR6oVCjA=",
+      "dev": true
+    },
+    "abbrev": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.0.tgz",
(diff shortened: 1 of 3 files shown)


MarshallOfSound:
@jjerome00 Cheers for the PR, this feature already exists from the tray menu however.

`Menu --> Help --> About`

It will launch a dialog which contains exactly the information in this new tab 😄

MarshallOfSound:
> I'm rather new to React, but I was able to figure out the pattern from other parts of the code. I'm also rather new to open source PR's so I apologize in advance if I missed something.

FWIW you nailed to code 👍 💯

jjerome00:
Wow, I saw the `AboutModal` in the code but for the life of me I could not figure out where it was in the app.  I'd still recommend adding the version somewhere within the main application window, but I won't push it.  Thanks for the feedback!

MarshallOfSound:
> I'd still recommend adding the version somewhere within the main application window

Where would you suggest putting it, I've played with putting it in the player window before (instead of a modal) but wherever I put it, it just took up to much space 😆

jjerome00:
Originally I thought if there was an About tab, you could add whatever information you wanted.  This is why I just straight-up copied the AboutModal contents.

However, there are a few spots with a little bit of space:
* The `SettingsPage` header, justified on the right
* The bottom of the hamburger menu, with a light font

I've attached a couple of very rough examples.  Maybe they'll give you some ideas.  Thanks for taking the time to listen!

settings dialog with version

hamburger-menu

PR #2793, opened 2017-10-01
Disable double-tap to zoom.

### LINKED ISSUE #2592: Enhacement request: disable Double-tap zoom
When double-tapping (mostly by accident) in the player window the screen zooms in as it is common browser behaviour.
There is no easy way to revert this zoomed in version though, because Cmd+0 doesn;t work and there is no "reset page zoom" option in the View menu.

Please disable the double-tap zoom or enable some options to reset it back to normal zoom.

**OS:**
macOS Sierra

**Issue Descriptions:**
See Above

**Steps to Reproduce:**
- double tap in main window
- the UI zooms in
- try to easily zoom back out

KarboniteKream:
Fixes #2592.

### DIFF
--- a/src/renderer/generic/index.js
+++ b/src/renderer/generic/index.js
@@ -1,4 +1,4 @@
-import { remote } from 'electron';
+import { remote, webFrame } from 'electron';
 import '../rendererEmitter';
 import './core';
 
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
Seems legit 👍

PR #2795, opened 2017-10-01
Remove default 'Recents' sidebar button

### LINKED ISSUE #2463: Recents button has no use on Free GPM
**OS:**
Win 10
**Issue Descriptions:**
The Recents Button does nothing on the Free version of GPM as the Home button does the exact same thing... so upon clicking the Recents it just takes the user to the home tab.
**Steps to Reproduce:**
- Have a free Google Play music account
- Click on the Recents button.

**Idea**
Make a toggle within the Settings!

KarboniteKream:
The sidebar and the side menu seems to be populated by GPM. Since the "Recents" button is added automatically if the user's subscription allows this feature, it's unnecessary to manually add the button (we also can't detect if the user is subscribed to GPM or not). Fixes #2463.

### DIFF
--- a/src/renderer/windows/GPMWebView/interface/customUI.js
+++ b/src/renderer/windows/GPMWebView/interface/customUI.js
@@ -132,15 +132,10 @@ function installAlarmButton() {
   });
 }
 
-function installRecentsButton() {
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
Haha, didn't actually notice Google added this link in themselves 👍

Issue #2675, opened 2017-07-29
Feature request: if application can't find lyric ask to open web browser with google search page where auto add artist and song name


Stigmaleon:
**OS:** Win 10

**Issue Descriptions:**
For some songs, especially for russian or some exotics)),  app can't find lyric. So it will be good add some button or some link to open google search with added artist and song name.

For example:
![](https://image.prntscr.com/image/m2dOE_PMQ96E8_IJIra2Sg.png)

dKab:
Working on it.

MarshallOfSound:
Fixed in #2803

PR #2803, opened 2017-10-04
Add "search for lyrics in google" link if lyrics could not be found 

dKab:
Closes (#2675)
image

### DIFF
--- a/src/renderer/ui/components/generic/LyricsViewer.js
+++ b/src/renderer/ui/components/generic/LyricsViewer.js
@@ -1,7 +1,7 @@
 import $ from 'jquery';
 import React, { Component, PropTypes } from 'react';
 import { findDOMNode } from 'react-dom';
-
+import { shell, remote } from 'electron';
 import { requireSettings } from './SettingsProvider';
 
 class LyricsViewer extends Component {
@@ -11,11 +11,21 @@ class LyricsViewer extends Component {
     themeType: PropTypes.string.isRequired,
   };
 
+  static openGoogleSearch(event) {
+    event.stopPropagation();
+    const { artist, title } = remote.getGlobal('PlaybackAPI').currentSong(true);
+    const lyricsTranslation = TranslationProvider.query('lyrics-lyrics');
+    const query = encodeURIComponent(`${title} - ${artist} ${lyricsTranslation}`);
+
+    shell.openExternal(`https://www.google.com/search?q=${query}`);
+  }
+
   constructor(...args) {
     super(...args);
 
     this.state = {
       visible: false,
+      noLyricsFound: false,
     };
   }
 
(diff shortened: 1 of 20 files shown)


src/renderer/ui/components/generic/LyricsViewer.js
```diff
@@ -1,21 +1,31 @@
 import $ from 'jquery';
 import React, { Component, PropTypes } from 'react';
 import { findDOMNode } from 'react-dom';
-
+import { shell, remote } from 'electron';
 import { requireSettings } from './SettingsProvider';
 
-class LyricsViewer extends Component {
+export class LyricsViewerComponent extends Component {
```
dKab: The default export is remaining untouched, so the modules that are using LyricsViewer won't break. This additional export of the component itself (not a HOC wrapped around it) is required in order to be able to call its method in test suite. To call class methods of component rendered with enzyme we need to call `component.instance()` to get the instance of its class. If we import the HOC exported by default, the call to `component.instance()` returns wrong object which doesn't have methods we want to test.

MarshallOfSound: I'd rather we don't test a component that isn't technically being rendered directly.  You can fetch the inner component to call it's method using the enzyme `find` method.

https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/find.md

test/electron-renderer/ui/LyricsViewer_spec.js
```diff
@@ -1,12 +1,14 @@
-/* eslint-disable no-unused-expressions */
+/* eslint-disable no-unused-expressions, no-duplicate-imports, import/no-duplicates */
```
dKab: some of the tests are actually expecting HOC wrapper which is being exported by default and won't pass if we use pure component class. The new tests require the unwrapped component itself in order to call its methods directly. So, I had to turn off this rule for this file and import two things from one file.  The other option is not to turn off eslint but instead create separate file for HOC wrapper.

dKab: Some of the tests are actually expecting HOC wrapper which is being exported by default and won't pass if we use pure component class. The new tests require the unwrapped component itself in order to call its methods directly. So, I had to turn off this rule for this file and import two things from one file.  The other option is not to turn off eslint but instead create separate file for HOC wrapper.

src/renderer/ui/components/generic/LyricsViewer.js
dKab: Got it, will try `find`

dKab:
@MarshallOfSound the solution with `find` method that you suggested didn't work for some reason:
image

But, I've found a way to test this indirectly (which is even better, to my mind). I just wait till `handleLyricsNotFound()` will be called automatically when 4 seconds have passed.  Sinon has a handy method for emulating passing of time which allows tests to be fast. I also changed one existing test to use this technique. Now tests run time has dropped by 4.5 seconds and we don't need to export underlying component after all.

MarshallOfSound:
> Now tests run time has dropped by 4.5 seconds and we don't need to export underlying component after all.

Nice!!! 👍

MarshallOfSound:
Thanks for this 😄

sayoojsamuel:
The `Search on google.com` option often searches with the query 
```
*<song name> - Unknown Artist lyrics*
```
This often result in a Google search without any Rich Text lyrics panel.  In the case of unknown artist, just a *<song name> lyrics* query would suffice, with better chance for formatted lyrics.

PR #2825, opened 2017-10-14
Include Space as an action key

linghutiantian:
As suggested by Samuel Attard, adding space as an action key, so we can now use global hotkeys including space, e.g. "ctrl+alt+space" etc.


linghutiantian:
A gentle ping~

PR #2872, opened 2017-11-19
Use HTTP for lastfm calls

metalmatze:
I just noticed in Wireshark that all requests send to LastFM are made with http although https is supported and their default.
I didn't test this change, but it should be an obvious one. :blush:

### DIFF
--- a/src/main/features/core/lastFM.js
+++ b/src/main/features/core/lastFM.js
@@ -7,6 +7,7 @@ import { LASTFM_API_KEY, LASTFM_API_SECRET } from '../../constants';
 const lastfm = new LastFmNode({
   api_key: LASTFM_API_KEY,
   secret: LASTFM_API_SECRET,
+  host: 'https://ws.audioscrobbler.com',
   useragent: 'GPMDP',
 });
 


MarshallOfSound:
Good catch, cant believe the API module would default to insecure...

I'll just merge this and test it manually later :)

jostrander:
I'm not sure that it will work unfortunately, it looks like they hard code the API calls to port 80 here:

https://github.com/jammus/lastfm-node/blob/master/lib/lastfm/lastfm-request.js#L42

MarshallOfSound:
:rip:. Reverted till we find a way around the hardcoded port

jostrander:
Honestly, the last commit to that library was over 3 years ago. Should we just fork and keep a copy on the gpmdp group?

MarshallOfSound:
I'd be down for publishing a fork based on the PR that fixes that issue.  Hopefully it'll get merged in upstream and we can drop the fork but it doesn't look like the maintainer is very active

jostrander:
[Forked](https://github.com/GPMDP/lastfm-node). I think we need to change a few more things to make it full proof.

metalmatze:
This is amazing. Thank you!

MarshallOfSound:
@jostrander Do you have permissions to publish to `@gpmdp` on NPM?

jostrander:
I believe so

jostrander:
Scratch that, I do not.

MarshallOfSound:
@jostrander Now you do 👍

PR #2883, opened 2017-11-28
Add a clear, obvious way to remove hotkeys to improve UX

zbrcz:
The solution is based on bradrisse's comment here:
https://github.com/mui-org/material-ui/issues/2932

Based on that thread, it seems to me that there is currently no better way to add the requested button.

### DIFF
--- a/src/renderer/ui/components/settings/HotkeyInput.js
+++ b/src/renderer/ui/components/settings/HotkeyInput.js
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 import React, { Component, PropTypes } from 'react';
-import TextField from 'material-ui/TextField';
+import { FlatButton, TextField } from 'material-ui';
 
 import { requireSettings } from '../generic/SettingsProvider';
 import { ACCELERATOR_KEYS, MODIFIER_KEYS, ACTION_KEYS } from '../../utils/constants';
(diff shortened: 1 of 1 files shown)


src/renderer/ui/components/settings/HotkeyInput.js
```diff
@@ -10,6 +10,16 @@ const styles = {
     width: '50%',
     float: 'left',
     padding: '4px 6px',
+    position: 'relative',
+  },
+  clearButton: {
+    'background-color': 'Transparent',
```
MarshallOfSound: This can just be `backgroundColor` and the value should be all lower case 👍

zbrcz:
Please note that I have amended the commit. I did it using "git push -f", not sure if that's the right way. :)

The requested screenshot:
hotkeyremovebutton

jostrander:
Looks reasonable. Does it also work with the dark theme?

zbrcz:
No, the icon doesn't change color. Thanks for pointing that out, I'll take a look at it.

zbrcz:
I have changed the `<button>` to a `<FlatButton>` and the arrow now changes color according to the theme. How do I submit the change? I get the feeling that _git push -f_ is not the right way.

jostrander:
`git push -f` is the right way if you are squashing your commits every time, otherwise a regular `git push` would do.

jostrander:
@MarshallOfSound this is good to go, just pulled and reviewed.

Issue #2889, opened 2017-12-02
[Tip] [Linux] The the app icon themed


SolarLiner:
**OS:** Manjaro Linux / Ubuntu 17.10

**Issue Description:** In order to get the icon themed, we can fool the icon theme into thinking the app is actually the Google Play Music Chrome extension, which is themed by several icon themes (Arc-Maia on Manjaro, Papirus too).

**Steps to Reproduce:** On an editor with root privileges:
* Edit `/usr/share/applications/gpmdp.desktop`
* Replace the value of the 'Icon' key with `chrome-fahmaaghhglfmonjliepjlchgpgfmobi-Default` (the chrome extension icon 'name')

Example edited `.desktop` file:
```
[Desktop Entry]
Name=Google Play Music Desktop Player
Comment=An electron wrapper for Google Play Music
GenericName=Google Play Music Desktop Player
Exec=gpmdp %U
Icon=chrome-fahmaaghhglfmonjliepjlchgpgfmobi-Default
Type=Application
StartupNotify=true
Categories=AudioVideo;Player;Audio;
StartupWMClass=Google Play Music Desktop Player
```

Sorry if this doesn't belong here, but I thought some might appreciate this little tip on getting the GPM icon to fit in with their theme.

MarshallOfSound:
See #2890

PR #2890, opened 2017-12-02
Localized .desktop file in French

SolarLiner:
.desktop files can be localized by appending the locale code in brakets to the key name. This ensures better integration with localized Linux desktops.


SolarLiner:
capture du 2017-12-02 23-22-04
Example result of localization (+result of icon theming 'hack' in #2889 )

MarshallOfSound:
I'll take the localization changes but not the changes that affect `Exec` or the icon.  I don't think we should be pretending to be another app (even if it is just the icon), it can and probably will confuse people and lead to unexpected behavior for people who already theme our icon.

The `Exec` changes just look like trouble, the current setup works 😄

SolarLiner:
Cleaned the file by only adding localization from the `.desktop` file in master.

Also I notice GenericName contains branding, when it shouldn't (see the vague [FreeDesktop spec, table 2](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html)). I can poke you with another PR with that, with a one-line change only, this time ⚡️

PR #2925, opened 2017-12-24
Discord Rich Presence 

aidant:
I've added more information to the Rich Presence Integration. 

- Play/Pause State
- Queue Length & current song in queue (using group size and players)
- Fixed a bug causing `00:00 left` to be displayed for initial song

Playing State
Paused State

These changes do need a new image for paused state. 
- Type: small
- Name: pause
- Source: [Download from here](https://cdn.discordapp.com/attachments/324908432534798336/394363318703489025/unknown.png)

Known Issues: 
1. If a queue contains one song many times it will always select the last occurrence of the song.
2. `PLAYING` state instead of `LISTENING` this is a limit of RPC and cant be changed yet.
3. If a user is idle or the queue is finished, the paused state will still be active. A potential solution would be to have the RPC interaction on a child process that can be killed and therefor the Rich Presence integration can toggled on and off.

### DIFF
--- a/src/main/features/core/discordRickPresence.js
+++ b/src/main/features/core/discordRickPresence.js
@@ -56,7 +56,7 @@ const setPresence = () => {
     };
 
     if (queueTrackIndex > queueLength || !queueTrackIndex) {
-      delete presence.pratySize;
+      delete presence.partySize;
       delete presence.partyMax;
     }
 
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
I added that small image so should be good on my app ID now 👍 

> If a user is idle or the queue is finished, the paused state will still be active. A potential solution would be to have the RPC interaction on a child process that can be killed and therefor the Rich Presence integration can toggled on and off.

Sounds like something worth exploring (forking the Electron main process isn't too hard to do 😄 )

PR #2976, opened 2018-02-01
Update node-engine in snapcraft.yaml

malept:
discord.js requires node 8, which is probably why 4.5.0 isn't in the Snap Store yet.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -36,7 +36,7 @@ parts:
 
   google-play-music-desktop-player:
     plugin: gulp
-    node-engine: 7.2.1
+    node-engine: 8.9.4
     source: .
     source-type: git
     gulp-tasks:


jostrander:
Can you show where discord requires node 8? I don't believe that's the case.

malept:
https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/60649402bd2ee91e3766c31530df3e0051e9f2b5/yarn.lock#L1930-L1938

Corresponding discord.js:

https://github.com/discordjs/discord.js/blob/d46eec4da40c01d134d863b9421fbc12d2ec6d20/package.json#L59-L61

jostrander:
LGTM, verified. On another note does snap craft auto update the store? I believe @RAOF set up the snapcraft package.

RAOF:
The snap in the `edge` channel is automatically updated. It needs someone to manually promote it from there.

Issue #2978, opened 2018-02-01
Discord Rich Presence layout could be improved


bramhaag:
The current layout looks a bit weird in my opinion:
```
Google Play Music Desktop Player
Album {album} by {artists}
{song} ({x} out of {y})
xx:xx left
```

The `Album {album} by {artists}` line is in my opinion very confusing, the `{artists}` are the artists of the current song playing, not the "artist" of the album, i.e albums such as Super Eurobeat have songs from 10 or more artists. In my opinion, the layout could be improved by either making it fully customizable or by using something like [Discord's Spotify layout](https://owo.whats-th.is/34dd8b.png) (I am aware that this uses some restricted features)

tankerkiller125:
I would like to see the format to be more like that of the Spotify Rich Presence

![Spotify Rich Presence](https://cdn-03.anonfile.com/99354bd0be/89d7af92-1517514377/Discord-Spotify-796x416.jpg)

mrpotatofactory:
You should contact the Discord Team about "advanced features".
You probably could contact the Discord Team, you most likely would get the same treatment as Spotify did hopefully without royalties.

![features](https://img.synergyst.club/@potatofactory/chrome_2018-02-01_15-57-46.png)

jostrander:
The Spotify integration appears to be a custom integration not using their standard RPC. I'm not sure that the advanced rich presence features would do any good in this case without contacting Discord for a similar layout.

jostrander:
Also, @bramhaag the track artist is currently what's selected. We can only pull that data from the currently playing song area, so if that's what Google displays there our hands are tied.

https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/master/src/main/features/core/discordRichPresence.js#L51

Alipoodle:
#2970 I've already mentioned the problem with it being cluttered, and even refereed back to the Discord Wiki... And still feel that some options should be made to make it seem less cluttered, especially with the amount of _useless_ information that is thrown at the user.

Stuff like ``Listening to Google Play Music Desktop Player`` and Album Art Is not possible due to using the Rich Presences API which only meant to be supported for Games (therefore playing) and images shouldn't be made on the fly / grabbed from another source..

[...]

PR #2991, opened 2018-02-07
Remove album from details in Discord Rich Presence

clavierpaul:
The album name is arguably the least important part of the song description, yet in the current implementation it comes before artist. This often causes text clipping in discord, hiding the artist name whilst showing the album it's on.

This also works better for singles, as the name of the song is duplicated in the current implementation (as album name == song name). Overall, it's simply a cleaner way of displaying song info.

A surprisingly common example below:
image

Notice how Everchanging is duplicated top and bottom, and Haywyre (the artist) is only visible by the letter "H"

Instead, why not display it like this?
image

### DIFF
--- a/src/main/features/core/discordRichPresence.js
+++ b/src/main/features/core/discordRichPresence.js
@@ -48,7 +48,7 @@ const setPresence = () => {
 
     const presence = {
       state: track.title,
-      details: `Album ${track.album} by ${track.artist}`,
+      details: track.artist,
       startTimestamp: start,
       endTimestamp: end,
       instance: false,


jostrander:
Personally I prefer this view of the existing one, so LGTM. It seems to be the general consensus in #2978 as well.

Evanito:
When is this going to be put into a release?

simoniz0r:
I'd be up for a new release just for this, tbh lol... the current layout is kinda not great haha

clavierpaul:
If you download one of the build artifacts the change is there, that's what I've done.

Issue #2972, opened 2018-01-31
Snap build for 4.5.0 is missing on the snapstore


Bohne13:
I can't find someone who supports the snap package on snapstore. The snap version for 4.5.0 should be updated if it works as a snap.

jostrander:
@RAOF What's the status on this?

Bohne13:
I have tried to start the app after upgrading do the --edge snap channel. But the app doesn't start!

jostrander:
Any error messages?

Bohne13:
Nope, it only don't want to load the aplication.

RAOF:
Hrm. Why didn't I see this?

Something screwy is going on with the naming of the binaries - it's coming through as `'Google Play Music Desktop Player'`, which is not what the launcher is expecting.

Let's see what's broken…

Feichtmeier:
Hi - just wondering - why is the app only availible in 4.4.1 version in the snap store if not installed via edge?

PR #2992, opened 2018-02-08
snap: Fix command escaping. (Fixes: #2972)

RAOF:
The snap now actually runs!


jostrander:
Out of curiosity, what changed that caused the old syntax to not work?

RAOF:
I *think* it was a change in the desktop-launcher part, changing the exec line from not quoting to quoting, so where we had whitespace escaped with `\ ` that then got passed through as a *literal* `\ `.

PR #3001, opened 2018-02-11
WIP: Possible fix for discord rich presence memory leak

jostrander:
I tested this on my machine and it seems to fix the memory leak, however removing this code also doesn't cause the memory leak on my machine, so I'm contributing it as a possible fix although I can't test it anymore.


MarshallOfSound:
Will leave the actual issue open until I have to validate this fix but it looks good to me

jostrander:
Sorry for the massive CI backup, was trying to get my commit to show up as signed.

PR #3019, opened 2018-02-20
Add Gentoo OS support in readme

Atoms:
Add Gentoo OS support in Readme 
for installation on Gentoo use eselect to enable repository rukruk (`eselect repository enable rukruk`) and then `emerge media-sound/google-play-music-desktop-player-bin`

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ OS Support
 * Mint
 * Fedora
 * Debian
+* [Gentoo](https://github.com/Atoms/rukruk) from Overlay
 
 Features
 --------


jostrander:
I'm a little hesitant to add this under OS support as we will inevitably get issues filed for Gentoo that we have no idea how to fix. Maybe the wiki would be a better spot to add this?

jostrander:
Definitely not opposed to adding a section for Gentoo, Arch and the SNAP Builds that's separate from the supported OS area. :+1:

BrianAllred:
Oh, I meant in the wiki, sorry I wasn't clear. Although linking to the Unofficial Releases wiki page on the readme would be cool. That way users have easy access to the information, but it's on a separate page where disclaimers and such could be made without cluttering up the readme.

jostrander:
Yeah I realized what you meant. We can link to "other releases" in the Wiki from the readme for the third party maintained stuff, including links to AUR etc. I believe anyone can edit the wiki, correct?

Atoms:
Guys i would not call those as releases, cause that's just packaging for different OS :)

BrianAllred:
Page is up [here](https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/wiki/Unofficial-Releases). @jostrander is right, anyone can edit the wiki, so if I screwed something up, feel free to fix it. 😄

Atoms:
seems ok from my side, thank you.

PR #3023, opened 2018-02-26
Intercept and fake a communication channel to the GPM chrome extension

### LINKED ISSUE #3018: Player stops sending track position information when minimized
**OS:** Windows 10 1703

**Issue Descriptions:** When the player is minimized (both to tray and just to taskbar) it stops sending track position (elapsed time) through the WebSocket API. When the player is brought back to the screen, the position updates as normal, and it updates properly from a second virtual desktop as well.

**Steps to Reproduce:** Use a program that hooks into the WebSocket API, play a track, and minimize the player. The track position fails to update until the player is brought back to the window.

MarshallOfSound:
Fixes #3018

This took a pretty long time to figure out (think weeks of haxing) but this should fix everyones minimized issues.  These events are always emitted regardless of background state (doesn't touch `requestAnimationFrame`) so this should be a good solution going forward

/cc @jostrander could you please test this 😄


jostrander:
Fantastic work, works beautifully! No idea why I didn't think of this.

jostrander:
Progress? :smile:

PR #3024, opened 2018-02-26
Changed update URL as xyz TLD was being blocked on my work PC

MikeKlebolt:
Feel free to use this PR or close it.  I made the change for myself as it appears GPMDP is checking for updates about every 2 minutes 15 seconds.  This was causing me to get spammed with notifications from Symantec saying that a URL was blocked.

 I may be the only one whose work is blocking the xyz TLD.

### DIFF
--- a/src/main/features/core/autoUpdater.js
+++ b/src/main/features/core/autoUpdater.js
@@ -9,7 +9,7 @@ if (process.platform === 'darwin') {
 const setUpAutoUpdate = () => {
   if (global.DEV_MODE) return;
   try {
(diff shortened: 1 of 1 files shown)


Als3mi:
Yahya

MarshallOfSound:
> I may be the only one whose work is blocking the xyz TLD.

I'm open to changing the update URL if this is a widespread issue, but not to `googleplaymusicdesktopplayer.com` as that domain has a *chance* that it could be revoked due to containing a trademarked term.  I can look into an alterante tld if required but afaik you're the first to have this issue 😀

PR #3063, opened 2018-03-24
Debounce forward/previous on the emitter to prevent double skipping

### LINKED ISSUE #2364: Make W10 Media Service actually usable
OS: Win10

Issue Descriptions: Media Keys skip twice if W10 Media Service is enabled.

Steps to Reproduce: Enable W10 Media Service, push an actual media key

According to previous Issues you implemented a "fix" so that media keys don't double-skip tracks. 

Apparently that fix was to disable the W10 Media Service. But I'd actually like to be using that, this seems more of a workaround then an actual fix. Because the error reappears if I use the W10 Media Service popup.

jostrander:
Maybe fixes #2364. Test before merging.

### DIFF
--- a/src/renderer/windows/GPMWebView/playback/controller.js
+++ b/src/renderer/windows/GPMWebView/playback/controller.js
@@ -17,9 +17,9 @@ window.wait(() => {
   );
 });
 
-Emitter.on('playback:playPause', () => {
-  window.GPM.playback.playPause();
-});
+Emitter.on('playback:playPause',
+  _.debounce(window.GPM.playback.playPause, 300, { leading: true })
+);
 
 Emitter.on('playback:play:smooth', () => {
   const originalVolume = window.GPM.volume.getVolume();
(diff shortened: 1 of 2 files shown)


BrianAllred:
Just curious, but wouldn't all button presses benefit from debouncing?

jostrander:
They probably would, however I didn't specifically have complaints about those so I figured "if it ain't broke don't fix it".

alok-sin:
@jostrander the play/pause also has this issue. But it's erratic. Perhaps debouncing will apply to that.

jostrander:
Can anyone with the issue at hand confirm this fixes it for double skipping before I implement play/pause as well?

MarshallOfSound:
@jostrander If you add play/pause I'll just merge it 😄

PickNChew:
Does not seem to work. Still double skips.

jostrander:
Added play/pause. As a side note, the pre-commit hooks don't work.

crstnio:
I applied this patch and play/pause doesn't work at all for me.

Having an issue #3155 with Apple EarPods immediately starting play after pausing.

PR #3112, opened 2018-05-05
Slack integration to update status

shaynemartin91:
# Description

I use [Slack](https://slack.com/) at work as well as for several private groups. I previously wrote a small script that uses the playback api to update my slack status on some of my slack teams. Right now its just a script that runs locally but I wanted to see it integrated as an actual feature so I've created this PR.
 
I know the contribution guidlines suggest asking before adding features, but I already had the core of the functionality written out for my own purposes so this was mainly just adding the setting page.

# Usage

Updates the status emoji to the headphones emoji and status text to the song and artist.

screen shot 2018-05-05 at 2 14 31 pm

Status is reset when playback stops or when the app is closed.

screen shot 2018-05-05 at 2 15 46 pm

# Testing

I added a spec for the new setting page. However, I couldn't seem to find any pattern for testing core features. If I missed a spot to put those tests, I'd be happy to add unit tests for `slack.js`.

# Questions

It's common for slack users to be part of multiple teams. I'm in 9 teams and would really like for this to update my status on 4 of them. I couldn't find a pattern for dynamically adding settings or where the setting value is an array and the user can dynamically add new values. Alternatively, I considered letting the field be a comma-delimited list of tokens but that feels hacky from a user's perspective. Is there a way to achieve something like this?

# Todo

- [ ] Unit testing for `slack.js` if there is a convention/place to add it
- [ ] Adding entries to all of the locales (that part might actually probably take more time than the rest of this,  so I wanted to make sure this feature was acceptable before doing that part)

### DIFF
--- a/src/_locales/en-US.json
+++ b/src/_locales/en-US.json
@@ -132,6 +132,9 @@
   "settings-options-style-dialog-css-files": "CSS Files",
   "settings-options-style-dialog-all-files": "All Files",
 
+  "slack-token-label": "Enter your slack user token (can be found at https://api.slack.com/custom-integrations/legacy-tokens)",
+  "slack-token-placeholder": "Slack token",
+
   "title-color-picker": "Color Picker",
   "title-settings": "Settings",
   "title-settings-playback": "Playback",
@@ -141,6 +144,7 @@
   "title-settings-listenbrainz": "ListenBrainz",
   "title-settings-mini": "Mini Player",
   "title-settings-style": "Custom Styles",
+  "title-settings-slack": "Slack",
 
   "tray-label-audio-device": "Audio Device",
   "tray-label-show": "Show"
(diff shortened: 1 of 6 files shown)


MarshallOfSound:
>  I couldn't find a pattern for dynamically adding settings or where the setting value is an array and the user can dynamically add new values

I don't think we have an existing pattern for this but a good one, in general, is to store the keys in an array in settings, and in the UI just show one input box that adds things to this array.  Render all keys as uneditable text on the settings page with a "remove" icon next to them.  Then they can add / remove them as they wish.

Overall this feature looks pretty sick 👍

tadgh:
I wrote this same thing myself as a standalone python script, glad to see it's being considered. Anything I can do to help get this moving?

jordyvandomselaar:
I'd be thrilled if this feature is added, my coworkers will finally be able to see what I'm listening to instead of having to ask where that noise is coming from 😉️

Jokes aside, I really like this feature.

PR #3321, opened 2018-08-08
Make thumbs up/down options in the tray menu work as checkboxes

calvin-sykes:
This PR makes the thumbs up/down options in the system tray work like checkboxes rather than plain buttons. 

So when a track is played, if it has already been rated the appropriate option in the tray will be shown checked. Changing the rating or setting one for the first time, either through the tray or through the main player window, updates the other properly (so you can unrate a track through the tray which wasn't possible before).

### DIFF
--- a/src/renderer/windows/GPMWebView/playback/controller.js
+++ b/src/renderer/windows/GPMWebView/playback/controller.js
@@ -45,20 +45,28 @@ Emitter.on('playback:stop', () => {
 
 Emitter.on('playback:thumbsUp', () => {
   if (!remote.getGlobal('PlaybackAPI').data.song.title) return;
-  new Notification('You just liked', { // eslint-disable-line
-    body: remote.getGlobal('PlaybackAPI').data.song.title,
-    icon: remote.getGlobal('PlaybackAPI').data.song.albumArt,
-  });
-  window.GPM.rating.setRating(5);
+  if (window.GPM.rating.getRating() !== '5') {
+    new Notification('You just liked', { // eslint-disable-line
+      body: remote.getGlobal('PlaybackAPI').data.song.title,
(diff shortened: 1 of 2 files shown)


src/renderer/windows/GPMWebView/playback/controller.js
```diff
@@ -45,20 +45,28 @@ Emitter.on('playback:stop', () => {
 
 Emitter.on('playback:thumbsUp', () => {
   if (!remote.getGlobal('PlaybackAPI').data.song.title) return;
-  new Notification('You just liked', { // eslint-disable-line
-    body: remote.getGlobal('PlaybackAPI').data.song.title,
-    icon: remote.getGlobal('PlaybackAPI').data.song.albumArt,
-  });
-  window.GPM.rating.setRating(5);
+  if (window.GPM.rating.getRating() !== '5') {
```
MarshallOfSound: Not 100% sure how this change is relevent or what it does?  Can you either clarify or remove ❓

calvin-sykes: Before, the toast got fired every time the menu option was clicked - which was fine because once a song was already liked clicking the button again didn't do anything.
The menu option is now a toggle, so depending on whether the song is currently liked or not the rating either needs to be reset ("un-liked", as distinct from disliked) or set to 5.
Only in the latter case should the "you just liked..." toast be shown. I didn't add an "un-liking" toast because I wasn't sure if that was the correct term to use.

(same argument in reverse applies to the dislike option)

Creepypoke:
Maybe, will be better, if you use radio buttons for this? It's more semantic correctly, also, if you group them together by the same `name` attribute, then browser will manage radiobutton state out of the box

PR #3378, opened 2018-10-08
Small change to make it show YouTube Music in Discord Rich presence

WalshyDev:
The large image should also be changed, I would need to know what the key is though before I can commit that.


MarshallOfSound:
IMO the image shouldn't be changed, it should be the logo of GPMDP.  If / when we re-brand towards something more generic or something YTM focused we can update that image 👍

Issue #3293, opened 2018-07-26
[4.6.1] Windows Multimedia Keys


jcursiolf:
**OS:** Windows 10 Home 64-bit

**GPMDP Version:** 4.6.1

**Issue Descriptions:** Multimedia track navigation keys misbehaving.

**Steps to Reproduce:** Pressing multimedia key forward on the keyboard will result in two tracks skipped. Pressing multimedia key backward on the keyboard will result in two tracks skipped. Example: assume you're listening to track 4 of a playlist/album. Pressing forward will go to track 6 instead of 5.

**Note:** Similar to closed [#2107](https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/issues/2107) from version 4.1.1.

Techman-:
I've come across this issue as well. It's especially prominent when I'm playing a video game (fullscreen).

amicloud:
I am unable to reproduce this issue. The multimedia keys work exactly as expected. On Win 10 64, GPMDP 4.6.1 as well.

Klorkson:
I, too, have this issue on WIndwos 10.
If it matters, my keyboard is the Logitech G910.
Also, in GPMDP (using YouTube Premium Music), I have the Windows 10 System Media Service (with song information) enabled.
What I have also noticed is that sometimes, when skipping songs, the playback gets paused.
It also happens both when the program is open in the fore gorund or in the back ground.

jcursiolf:
Adding a little bit more information: I've reinstalled and it initially did not present the problem. However, I remembered that I like to have the song info on the "Windows 10 Volume Control" and "Windows 10 Media Server". Enabling the Windows 10 Media Server option brought the bug back.  Disabling it made the bug go away again.

Techman-:
Yeah, it seems that disabling the Win10 media service gets rid of the issue, but it also gets rid of the nice volume overlay.

Alipoodle:
Double of: #2364
This is mainly due to Windows Media Service + GPMDP Still both picking up on the Next Media Key it seems.

Techman-:
With the Windows 10 media service disabled in settings, how do I get GPMDP to capture the media key events so I can control it while in a fullscreen game?

NetkoNefarious:
I have this same issue as well.

aaldrich29:
[...]

PR #3385, opened 2018-10-15
Fix Issue #3293

derpmarine168:


MarshallOfSound:
This completely disabled these hotkeys when the media service is active which as discussed through the issues in this repo might work for some people but not for all people.  E.g. This breaks it completely on my machine

MarshallOfSound:
I'd rather follow up on @jostrander 's PR / workaround

derpmarine168:
This fix works fine for me. The media keys should work with windows media
servics. If not, it's a bug with windows. (I'm runing windows 10 home)
DerpMarine | Website <http://DerpMarine168.github.io> | Send Me a Email
<derpmarine168@gmail.com>

PR #3505, opened 2019-02-27
Support for macbook touchbar

th3dark0n3:
This adds support for touch bar in MacBook

Tested on MacBook Pro 2018

img_20190227_122422

---
This change is [Reviewable](https://reviewable.io/reviews/marshallofsound/google-play-music-desktop-player-unofficial-/3505)


th3dark0n3:
Forgot to raise a feature request first, should I still raise it?

MarshallOfSound:
@th3dark0n3 Can you remove the slider (it doesn't look that great IMO and the native slider you get from `electron-media-sevice` is way better.  Also can you update the thumbs up / down buttons to be the material icons for thumbs up / down instead of the emojis 👍

th3dark0n3:
@MarshallOfSound changed, implemented like/dislike similar to app, i.e., fill when liked
Not sure what to do about the codeclimate issue, the code looks more readable when keeping the changes

th3dark0n3:
Yup I saw the PR in Electron
Let me know if I need to update this PR?

arongeorgel:
Will this contain the changes for integrating the media button while one is in another app like Spotify has it? (check images bellow)
IMG_20190412_144752
IMG_20190412_144808

PR #3513, opened 2019-03-05
Update to mpris-service 2.0.0

### LINKED ISSUE #2741: D-Bus wrong volume output
When the volume is at exactly 0% or 100%, the following command:

    qdbus org.mpris.MediaPlayer2.google-play-music-desktop-player \
          /org/mpris/MediaPlayer2 \
          org.mpris.MediaPlayer2.Player.Volume

outputs the raw byte 0x01:

    

Expected behavior is of course either the string 0 or 1, respectively.

All other numbers between 0 and 1 are outputted correctly.

acrisci:
Update the mpris-service dependency to the latest version. This update
should fix many of the outstanding issues with the GPMDP mpris
implementation.

fixes #2741

### DIFF
--- a/src/main/features/linux/mprisService.js
+++ b/src/main/features/linux/mprisService.js
@@ -1,20 +1,23 @@
 import { app } from 'electron';
-import _ from 'lodash';
 import mpris from 'mpris-service';
 
 function mprisService() {
   const mainWindow = WindowManager.getAll('main')[0];
-  let _songInfo = {};
 
   const player = mpris({
-    name: 'google-play-music-desktop-player',
+    name: 'google_play_music_desktop_player',
     identity: 'Google Play Music Desktop Player',
     canRaise: true,
     supportedInterfaces: ['player'],
     desktopEntry: 'google-play-music-desktop-player',
   });
+
   player.playbackStatus = 'Stopped';
   player.canEditTracks = false;
+
+  player.getPosition = function getPosition() {
+    return PlaybackAPI.currentTime().current * 1e3;
+  };
 
   player.on('raise', () => {
     mainWindow.setSkipTaskbar(false);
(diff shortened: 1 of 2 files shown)


acrisci:
Hello,

Many of my users on [Playerctl](https://github.com/acrisci/playerctl) have been having problems with players that use `mpris-service` for their implementation of mpris for desktop integration.

* https://github.com/acrisci/playerctl/issues/40
* https://github.com/acrisci/playerctl/issues/81
* https://github.com/acrisci/playerctl/issues/35

To fix this error, I have [rewritten the mpris-service library](https://github.com/emersion/mpris-service/releases/tag/v2.0.0) using a [dbus library](https://github.com/acrisci/node-dbus-next) that I wrote for this purpose that is a fork of [dbus-native](https://github.com/sidorares/dbus-native) with added features required for implementing mpris.

I have tested this branch with Playerctl and things work as expected.

There is a bit more feature work required to be fully compatible which I plan to do such as implementing the shuffle and loop status features.

Thanks.

src/main/features/linux/mprisService.js
```diff
@@ -7,15 +7,20 @@ function mprisService() {
   let _songInfo = {};
 
   const player = mpris({
-    name: 'google-play-music-desktop-player',
+    name: 'google_play_music_desktop_player',
```
acrisci: I was under the impression that dashes were not allowed in bus names, but I just looked this up and they are [allowed but discouraged](https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-bus). So I can change it back to what it was.

jostrander:
Testing...

jostrander:
Getting an error with the dbus interface name: 

```
error: Uncaught Exception. Error: Invalid interface name: org.mpris.MediaPlayer2.google-play-music-desktop-player
```
Appears that `-` is [not allowed](https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-interface) in the interface name, so we should probably switch to `_` for that as well.

jostrander:
False alarm, I forgot to build.

acrisci:
Thanks for the quick reply.

I've taken another look at it and implemented that position stuff in a force push.

Ready for another look.

PR #3524, opened 2019-03-13
Adding show option and shortcut to app menu.

robbyfletcher:
Adding Show option with shortcut to be able to bring application back open after it has been backgrounded.

### DIFF
--- a/src/main/features/core/applicationMenu.js
+++ b/src/main/features/core/applicationMenu.js
@@ -1,6 +1,8 @@
 import { app, Menu, shell } from 'electron';
 import { showDesktopSettings } from './desktopSettings';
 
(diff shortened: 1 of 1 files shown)


src/main/features/core/applicationMenu.js
```diff
@@ -1,6 +1,8 @@
 import { app, Menu, shell } from 'electron';
 import { showDesktopSettings } from './desktopSettings';
 
+const mainWindow = WindowManager.getAll('main')[0];
```
MarshallOfSound: There is no guaruntee that this is the same window or that it isn't `null`.  Can we use `WindowManager.getAll` inside the `click` handler and check that it isn't null before using it

PR #3530, opened 2019-03-18
Improve Japanese translation

yuk7:
* Default devices are generally translated as "既定のデバイス".(ex.Windows)
* Even in Japan, the brand "Youtube Music" remains the same.
* The notation "イシュー" is not common in Japan.


MarshallOfSound:
Thanks 👍

PR #3550, opened 2019-04-23
Set GTK dark theme if dark theme is enabled

### LINKED ISSUE #3541: Set dark window borders with dark theme on Linux
**Checklist:**

* [x] I have checked that there are no issues with similar or the same content

  **YOU SHOULD CHECK CLOSED ISSUES ASWELL**
* [x] I have checked the FAQ (https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/wiki/FAQ) and the answer I am looking for is not there
* [x] I have double checked and can reproduce the issue

**OS:** Linux ([elementary OS](https://elementary.io) 5.0 Juno)

**GPMDP Version:** 4.6.1

**Issue Descriptions:**

On GTK-based Linux desktops (like GNOME, XFCE, elementary OS, Ubuntu, etc.), apps can request a dark style. When they do so, they get a dark titlebar—for example, [Etcher](https://github.com/balena-io/etcher) (another Electron app) successfully does this to better match its UI. Some Atom extensions exist to do the same. Google Play Music Desktop Player could similarly set the dark style for its window borders to better blend in on Linux desktops.

Screenshot from 2019-04-17 01 03 56@2x

**Steps to Reproduce:**

1. Use a Linux-based OS with a light theme by default (like elementary OS)
2. Enable the dark theme in GPMDP
3. See the mis-matched window borders

cassidyjames:
Should fix #3541, but I have yet to test.

### DIFF
--- a/src/main/configureBrowser.js
+++ b/src/main/configureBrowser.js
@@ -18,6 +18,7 @@ export default () => {
     autoHideMenuBar: true,
     frame: Settings.get('nativeFrame'),
     titleBarStyle: Settings.get('nativeFrame') && process.platform === 'darwin' ? 'hidden' : 'default',
+    darkTheme: Settings.get('themeType') === 'FULL',
     icon: path.resolve(`${__dirname}/../assets/img/main.${(process.platform === 'win32' ? 'ico' : 'png')}`), // eslint-disable-line
     title: 'Google Play Music Desktop Player',
     webPreferences: {


cassidyjames:
I can confirm this works when the window opens. However it does not set it live, i.e. you have to close and re-open the window for it to apply.

kgrubb:
all windows should now follow the correct titlebar theme. some pics can be seen here: https://github.com/cassidyjames/Google-Play-Music-Desktop-Player-UNOFFICIAL-/pull/1#issuecomment-506009311

As for reloading browserWindow, does anyone have a good reference to a project that auto-reloads the browserWindow when titlebars are changed/modified? I think vscode does something like this, so I'll go see how they handle it.

PR #3551, opened 2019-04-23
Add `libavahi-compat-libdnssd-dev` to Linux dev requirements

cassidyjames:
This is needed to resolve an error about missing `dns_sd.h` on Ubuntu 18.04 and elementary OS 5 Juno.


jostrander:
This should be added to the wiki, if it isn't already. There's a development FAQ there with all the linux dependencies.

PR #3569, opened 2019-05-24
Update linux media key implementation to use dbus-next

### LINKED ISSUE #2864: MediaKeys support needs to be updated for GNOME 3.24+
Currently `org.gnome.SettingsDaemon` bus name is used: https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/e1437f2eccfc0d9ccf9619c85dbea27e0d22ab1e/src/main/features/linux/mediaKeysDBus.js#L4

This was renamed to `org.gnome.SettingsDaemon.MediaKeys` in 3.24 and the old name was removed in 3.26. This is noted by upstream here: https://mail.gnome.org/archives/desktop-devel-list/2017-April/msg00069.html

The solution is to just try both bus names for now.

jostrander:
Also fixes #2864

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,6 +161,7 @@
     "@nodert-win10/windows.storage.streams": "^0.2.95",
     "appdmg": "^0.4.5",
     "dbus": "GPMDP/node-dbus#linux-only",
+    "dbus-next": "^0.6.1",
     "electron-installer-debian": "^0.4.0",
     "electron-installer-redhat": "^0.3.0",
     "electron-media-service": "^0.2.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -161,6 +161,7 @@
     "@nodert-win10/windows.storage.streams": "^0.2.95",
     "appdmg": "^0.4.5",
     "dbus": "GPMDP/node-dbus#linux-only",
```
MarshallOfSound: Do we still need this `dbus` dep?

jostrander: Just checked and we do not. With the mpris change that got pulled in and now this it can be removed.

PR #3584, opened 2019-06-12
Update flora-colossus

jostrander:
Fixes modules in linux that crash on startup causing MPRIS and dbus services to fail.


jostrander:
transient error on appveyor build, could probably be retried and would work.

PR #3588, opened 2019-06-13
Upgrade to electron 4

jostrander:
There are quite a few changes with electron 4, and this serves as a starting point for that upgrade. I've fixed what appears to have been broken in electron 4's remote.getCurrentWindow() api returning null during the renderer pre script in linux.

Electron 4 [drops support for 32 bit](https://electronjs.org/blog/linux-32bit-support) packages in linux,~~but this hasn't removed it from the build tooling as of yet.~~

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -70,20 +70,6 @@ jobs:
             - "node_modules"
       - store_test_results:
           path: test_reports
-  linux_pack_32:
-    docker:
-      - image: gpmdp/build-32
-    steps:
-      - checkout
-      - restore_cache:
-          key: linux-modules-{{ .Branch }}-{{ .Revision }}
-      - run:
-          name: Package
-          command: yarn package:linux:32
-      - save_cache:
-          key: linux-packaged-32-{{ .Branch }}-{{ .Revision }}
-          paths:
-            - "dist"
   linux_pack_64:
     docker:
       - image: gpmdp/build-64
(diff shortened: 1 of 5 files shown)


jostrander:
Any chance you can pull and test this on windows / mac @MarshallOfSound ? I don't have access to either at the moment.

kgrubb:
hm, looks like the theming isn't working on this branch (macOS):
Screen Shot 2019-07-01 at 4 54 59 PM

For comparison, here's what it looks like on the master branch:
Screen Shot 2019-07-01 at 4 59 21 PM


My Mac Info, in case it helps anyone:
Screen Shot 2019-07-01 at 4 57 07 PM

jostrander:
@kgrubb Do you have any errors in the console? I don't have a mac to test on, so I can't see which modules are possibly broken for you.

Stanzilla:
@jostrander any plans to pick this up again? maybe even go to v6?

jostrander:
Yeah, I was trying to test on mac, but I'm having a problem I can't seem to solve as of yet. I would migrate to V6 but it's more involved than I want to get into at the moment.

jostrander:
Rebased this and upgraded to electron 4.2.12, which fixed the MacOS build issues.

reduckted:
I've been using this build for the last few days on Windows 10 and haven't had any issues with it.

jostrander:
@MarshallOfSound best I can tell this is ready to go, and is the furthest we can upgrade without rewriting native modules that won't compile in Node 12. I started to rewrite some native modules to compile for Electron 5+ but there are quite a few to do across platforms.

kgrubb:
Checking in to confirm, this seems to be working without any issues on macOS catalina as well (built with node 11).
Screen Shot 2019-12-09 at 12 30 09 PM

SolarAquarion:
I'm thinking that @MarshallOfSound is not particularly "active" on this project unfortunately? Because supporting the latest nodejs would be great

Issue #3545, opened 2019-04-19
Google Sign In may break in June 2019


MarshallOfSound:
Just raising this pre-emptively as a tracking issue for https://github.com/electron/electron/issues/17887

Unclear if whatever measures Google implement will affect Electron or if we can work around them locally.  If it's just UA sniffing we can defeat that easily, I guess we'll just have to wait and see what happens.

apt-itude:
I'm not sure if this is still being tracked, but as of the last few days, I can no longer use the app. It gets stuck at "Loading Music Library" indefinitely. This issue jumped out at me given the timing.

sulymkaa:
have the same issue, cant use desktop app on mac os, please fix asap

MarshallOfSound:
> It gets stuck at "Loading Music Library" indefinitely.

This was actually caused by an unrelated GPM incident.  I hit it yesterday on one of my machines, deleting your cache and relaunching will fix it.

andrevalgrande:
I'm stuck on `Loading Music Library…`, but I can't figure out how to delete the application cache on MacOS. I've tried killing the application, then wiping the two GPMDP folders under `~/Library/Caches`, but it didn't work. Am I missing something?

jfries289:
@MarshallOfSound I'm experiencing the same issue (OSX). I deleted the cache located at `~/Library/Caches/google-play-music-desktop-player`, but the issue persists. Is there something else I should remove?

Darktex:
This started happening to me too! I also deleted my cache at `~/Library/Caches/google-play-music-desktop-player` to no avail :(

cscuderi:
Chiming in. The issue also happened to me on OS X. Stuck at "Loading music library" :(

cscuderi:
I deleted the cache folder at: `/Users/carlinscuderi/Library/Application Support/Google Play Music Desktop Player/Cache`, restarted the application, and it now works.

Darktex:
I haven't literally done anything but it started working on its own lol

ervinb:
I'm having issues signing in on W10.

![Capture](https://user-images.githubusercontent.com/2212814/62786800-a4560280-bac3-11e9-9a57-59942af89c8d.PNG)

noizyboy:
Same issue here as @ervinb. 

[...]

PR #3608, opened 2019-08-26
Use a configurable user agent to circumvent the sign in issue

ervinb:
Fixes https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/issues/3545

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -104,9 +104,14 @@ app.setAppUserModelId('com.marshallofsound.gpmdp.core');
   // initialization and is ready to create browser windows.
   app.on('ready', () => {
     mainWindow = new BrowserWindow(generateBrowserConfig());
-    // Remove Electron from the user agent
-    const newUserAgent = mainWindow.webContents.getUserAgent()
-      .replace(/Electron\/.+? /g, '');
+      // Use the default user agent but remove Electron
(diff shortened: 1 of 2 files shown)


WillPresley:
For what it's worth, this seemed to break Chromecast support for me. The icon was missing, and so was the 'extra' Miniplayer button. Going back to the nightly build from June 12th brought everything back without an issue, just thought I'd report the issue.

eturk1:
where can we get a beta .exe for Windows to test out sign fix? :)

MarshallOfSound:
@WillPresley Good catch, I think we need the fake user agent to include the "Chrome" string that the Chrome browser has. Can you test that by modifying your local settings file?

WillPresley:
@MarshallOfSound That seemed to work. Copied my own UA from latest Chrome on Windows: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.132 Safari/537.36` then pulled out the Windows-specific parts to leave just this: `Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.132 Safari/537.36` and the Chromecast features and bigger miniplayer button are back to normal.

That-Marketing-Guy:
> Fixes #3545

How is this suppose to be a fix for a general user???

PR #3632, opened 2019-09-25
Support for Super(Windows) key in shortcuts

huhle:
Just added Super key to the list of accelerators, so win key can by also used as part of shortcut.
Seems like a nice feature IMO


huhle:
Hi, I am pretty sure that I didn't break the darwin build. Is there anything I can do with it? Maybe just try it again?

Issue #33, opened 2015-11-08
Double audio if you open two instances


MarshallOfSound:
If you open two instance of GPMDP and play music in one it stutters strangely.  I think we should only allow one instance of GPMDP open at a time...

PR #3645, opened 2019-10-16
fix media key mapping on Linux

davegallant:
Media keys were not mapped / functioning on Linux.

## CHANGES
- call `GrabMediaPlayerKeys` on the MediaKeys D-Bus interface rather than the `ProxyObject`. This seems to be aligned with the `dbus-next` readme: https://github.com/dbusjs/node-dbus-next#the-client-interface

Tested this on the Ubuntu 18.04 and media keys are now functioning.

### DIFF
--- a/src/main/features/linux/mediaKeysDBus.js
+++ b/src/main/features/linux/mediaKeysDBus.js
@@ -13,17 +13,17 @@ async function registerBindings(desktopEnv, session) {
   };
 
   const legacy = await session.getProxyObject(`org.${desktopEnv}.SettingsDaemon`, `/org/${desktopEnv}/SettingsDaemon/MediaKeys`);
-  legacy.getInterface(`org.${desktopEnv}.SettingsDaemon.MediaKeys`);
-  legacy.on('MediaPlayerKeyPressed', listener);
+  interfaceLegacy = legacy.getInterface(`org.${desktopEnv}.SettingsDaemon.MediaKeys`);
+  interfaceLegacy.on('MediaPlayerKeyPressed', listener);
   app.on('browser-window-focus', () => {
-    legacy.GrabMediaPlayerKeys('GPMDP', 0); // eslint-disable-line
+    interfaceLegacy.GrabMediaPlayerKeys('GPMDP', 0); // eslint-disable-line
   });
 
   const future = await session.getProxyObject(`org.${desktopEnv}.SettingsDaemon.MediaKeys`, `/org/${desktopEnv}/SettingsDaemon/MediaKeys`);
-  future.getInterface(`org.${desktopEnv}.SettingsDaemon.MediaKeys`);
-  future.on('MediaPlayerKeyPressed', listener);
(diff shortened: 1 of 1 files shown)


src/main/features/linux/mediaKeysDBus.js
```diff
@@ -13,17 +13,17 @@ async function registerBindings(desktopEnv, session) {
   };
 
   const legacy = await session.getProxyObject(`org.${desktopEnv}.SettingsDaemon`, `/org/${desktopEnv}/SettingsDaemon/MediaKeys`);
-  legacy.getInterface(`org.${desktopEnv}.SettingsDaemon.MediaKeys`);
-  legacy.on('MediaPlayerKeyPressed', listener);
+  interfaceLegacy = legacy.getInterface(`org.${desktopEnv}.SettingsDaemon.MediaKeys`);
```
jostrander: This should be defined with a const before it.

src/main/features/linux/mediaKeysDBus.js
```diff
@@ -13,17 +13,17 @@ async function registerBindings(desktopEnv, session) {
...
+  interfaceLegacy.on('MediaPlayerKeyPressed', listener);
   app.on('browser-window-focus', () => {
-    legacy.GrabMediaPlayerKeys('GPMDP', 0); // eslint-disable-line
+    interfaceLegacy.GrabMediaPlayerKeys('GPMDP', 0); // eslint-disable-line
   });
 
   const future = await session.getProxyObject(`org.${desktopEnv}.SettingsDaemon.MediaKeys`, `/org/${desktopEnv}/SettingsDaemon/MediaKeys`);
-  future.getInterface(`org.${desktopEnv}.SettingsDaemon.MediaKeys`);
-  future.on('MediaPlayerKeyPressed', listener);
+  interfaceFuture = future.getInterface(`org.${desktopEnv}.SettingsDaemon.MediaKeys`);
```
jostrander: Same here, should be `const interfaceFuture =`

davegallant: good point. done.

jostrander:
I pulled this to test, and it accepts a key once then no longer functions. Trying to make sure nothing else is the culprit.

davegallant:
> I pulled this to test, and it accepts a key once then no longer functions. Trying to make sure nothing else is the culprit.

@jostrander which Desktop Env are you using? I've had no issues on 2 machines running `Ubuntu 18.04` and `GNOME 3.28.4 `.

jostrander:
>>> 5.0.0-31-generic #33~18.04.1-Ubuntu SMP Tue Oct 1 10:20:39 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
gnome-shell                                3.28.4-0ubuntu18.04.2

jostrander:
This was an issue on my end. I had a bad set of node_modules from running `npm install` instead of `yarn`.

PR #3647, opened 2019-10-20
Add a button in each player to switch service

### LINKED ISSUE #3627: Accessibility of Youtube Music mode
It was very difficult to find the option to switch to Youtube Music in Ubuntu 18.04. I searched through the whole application, only to later find it in a dropdown in the newly appeared icon besides the accesibility icon on the Ubuntu Topbar. It would be better to also have a button to swtich to YT Music inside the application itself.

jostrander:
Fixes #3627 and maybe others.

image

image

### DIFF
--- a/src/_locales/cs.json
+++ b/src/_locales/cs.json
@@ -11,6 +11,8 @@
   "button-text-lets-do-this": "Pojďme do toho!",
   "button-text-lets-go": "Tak jdem na to!",
   "button-text-not-now": "Ne nyní",
+  "button-text-ytm-switch": "",
+  "button-text-gpm-switch": "",
 
   "label-about": "O Aplikaci",
   "label-alarm": "Budík",
--- a/src/_locales/da.json
+++ b/src/_locales/da.json
@@ -35,6 +35,8 @@
   "lastfm-login-not-authorized": "Dit login blev ikke godkendt",
   "lastfm-logout-button-text": "Log ud af Last.FM",
   "lastfm-map-thumbs-up-to-heart": "",
+  "button-text-ytm-switch": "",
+  "button-text-gpm-switch": "",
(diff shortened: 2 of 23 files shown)


Sk8erOfBodom:
I'm sorry, but this is horrible, at the very least please make it another color so it's not a big red button invading my app. Even Google isn't advertising Youtube Music in their Google Play Music applications.
I think it doesn't even make sense that Youtube Music is in GPMDP.

If you don't change anything, I'll still love GPMDP but I might like you less. 😛

Sk8erOfBodom:
> I'm sorry, but this is horrible, at the very least please make it another color so it's not a big red button invading my app. Even Google isn't advertising Youtube Music in their Google Play Music applications.
> I think it doesn't even make sense that Youtube Music is in GPMDP.
> 
> If you don't change anything, I'll still love GPMDP but I might like you less. 😛

Alright, I hid it myself with some CSS, but please still consider the issue, button with hard coded colors tarnishes one of the many great features of GPMDP!

wakamex:
what if I click the button once and I can't go back. was this a trap all along? where the heck is this setting saved, I tried deleting all the files and reinstalling. using manual install of Build #3838 (rpm/x64)on clear linux. it was working fine before i clicked the button. why did i have to click the button??
image

jostrander:
@wakamex That looks like linux or mac, try using the tray icon if you have one to switch services.

wakamex:
thanks for the quick reply. that looks like the right file. changed `"service": "youtube-music",` to `"service": "google-play-music",` and restarted. looks the exact same. check the file again, and it's just reverted back to youtube-music. so doesn't seem to be reading that in as input, only saving its internal state? or maybe I need to modify it differently.

PR #3661, opened 2019-11-01
Fix mpris support in Linux (tested with GNOME on GNU/Debian-sid).

### LINKED ISSUE #3618: MPRIS seems broken in KDE 5.16.5
**Checklist:**

* [x] I have checked that there are no issues with similar or the same content
  (I think similar in #3583 , but that issue in Gnome and not discussed for MPRIS specification.)

  **YOU SHOULD CHECK CLOSED ISSUES ASWELL**
* [x] I have checked the FAQ (https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/wiki/FAQ) and the answer I am looking for is not there
* [x] I have double checked and can reproduce the issue

**OS:**

Arch Linux (last updated in about 3:00 UTC)

**GPMDP Version:**

both:

- 4.6.1 from AUR
- 5332c05

**Issue Descriptions:**

After updating to KDE5.16.5, the media control applet no longer detects GPMDP.
According to KDE [changelog](https://kde.org/announcements/plasma-5.16.4-5.16.5-changelog.php), that I think requests more strict MPRIS specification.

> Players not compliant to MPRIS specification will no longer work with Media Controller applet. [Commit](https://cgit.kde.org/plasma-workspace.git/commit/?id=8eec2cfd36c4ecf9e2b13f4226ed49495271a1e2). Phabricator Code review D22596

**Steps to Reproduce:**

1. Play any music.
2. Media Controller applet has not reaction.
3. Sometimes not working media control keys without MPRIS.

TurboGit:
Fixes #3618.

### DIFF
--- a/package.json
+++ b/package.json
@@ -166,7 +166,7 @@
     "electron-media-service": "^0.2.2",
     "ll-keyboard-hook-win": "^3.0.0",
     "mouse-forward-back": "^1.0.1",
-    "mpris-service": "^2.1.0"
+    "mpris-service": "GPMDP/mpris-service"
   },
   "babel": {
     "presets": [


jostrander:
Can you explain why this fixes the mpris support? This just reverts recent changes that were made to bring mpris up to date.

TurboGit:
It reverts indeed a change that broke the MPRIS support. So reverting a broken patch looks like a fix to me? Just test it, it works. Note that only part of the patch is reverted.

I'm using this patch for my own build since some days on GNOME without issue.

jostrander:
The original change may have broken mpris support, but the current version  should also be compatible. I'd rather come to a solution with the new version vs reverting.

TurboGit:
But then the original author should look at this. I have mostly bisected the issue to the wrong commit and I have reverted the minimal part that broke completely the MSIS support. I'm no expert on this issue, I'm also contributing to too many Open Source projects and cannot invest more time on this. I wanted this to work as I depend on this issue and it was broken for far too long, so I decided to fix it. Feel free to close this PR if you don't think it can be integrated, that's fine with me.

jostrander:
I'm going to close this as it doesn't solve the issue without reverting way back before the dbus/mpris upgrade. If we can come up with a specific fix for the issue at hand otherwise I'd review it.

PR #3745, opened 2019-12-21
Update README.md

jnr2820:
Added note for Dev build search tool.
https://gpmdp.azurewebsites.net/
Project here: agc93/gpmdp-release

Thanks for taking the time to review this request
- jnr2820

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,9 @@ Download
 ---------
 Head over to our website http://www.googleplaymusicdesktopplayer.com to download the latest release for your platform.
 
+If you'd like to try out the latest "successful" development build, head over to this search tool: https://gpmdp.azurewebsites.net/ Project here: agc93/gpmdp-release
+Note: Though a build was successfully build, it may have severe defects. Use at your own risk.
+
 OS Support
 ------------
 


MarshallOfSound:
Hey @jnr2820 cool tool, looks like @agc93 built this so I'd like sign off from them that they're OK with this going in the README and I'd probably like to `CNAME` it to the `gpmdp` domain if they're OK with that 🤔

agc93:
Yep, that was me. The tool itself was always supposed to be a stop-gap to the problem of people having to use CI releases that were difficult to find. Ideally this wouldn't be the long-term solution through which users are installing GPMDP.

I'm totally okay with it going in the README, but given I'm still paying to host the service (currently) I'd prefer to either a) fully hand the tool over and allow you to host it somewhere under the GPMDP domain or b) leave it non-CNAME'd as-is (that is, a temporary solution).

Happy to work with either option, or any other approach you'd prefer 👍

Razzeee:
I guess the service backend is down by now? So the PR doesn't make sense anymore?

aokegbile (approved):

Issue #2687, opened 2017-08-04
Feature Request: Pause playback when machine is locked


eriqs:
**OS:**
Windows 10

**Issue Descriptions:**
Feature Request: Pause playback when machine is locked

**Steps to Reproduce:**
I love the app, thank you!  The one thing I don't like is that it doesn't have the option to pause playback when my computer is locked, like other similar apps I have used.  I use this at work, and always lock my computer when I walk away so I would like to have the music pause when I do this, and resume when I return - rather than needing to pause and resume it manually.  Thank you again for the excellent app!

jostrander:
There really is no formal API for this in electron as far as I can tell. So we can't reliably add this.

eriqs:
Bummer. Perhaps the other way around then?  As in a hot key that will pause the player and lock the machine simultaneously?

[sent from my mobile device]
On Aug 6, 2017, at 7:07 PM, Jesse Ostrander <notifications@github.com<mailto:notifications@github.com>> wrote:


There really is no formal API for this in electron as far as I can tell. So we can't reliably add this.

MarshallOfSound:
@jostrander It's not that there's no formal way to do it in Electron, I don't think there's a way to do it on all platforms.

I think `macOS` fires a sleep event, which doesn't exactly equal lock, and windows might fire a window message to the top level window we might be able to read, but on linux it's entirely up to the WM what to tell the windows and when 🤔 

Something to offload to a magical native module I feel, I'll take a look in chromium though and see if they've done the hard work for us.

flippinger:
Any way to do this via Windows task scheduler? At least there is a trigger "on workstation lock".

![image](https://user-images.githubusercontent.com/3148313/29188168-9039984c-7e12-11e7-8991-ea8cfe73a86b.png)

Kytech:
This can be implemented in Windows through the System Media Transport Controls, though we'd have to work out the double skip issue that is currently unresolved. (See #2253)

JasonLG1979:
On Linux it can be accomplished over DBus.

liamdawson:
> I think macOS fires a sleep event

High Sierra supports locking without sleep, so that method won't work in all cases.

IndefiniteBen:
@Kytech Yes please. Music control during connected standby is just about the only thing making me wish for a native Windows store version of GPM.

tuur29:
[...]

Issue #3598, opened 2019-07-21
Issue with media control buttons on Windows 10


eera5607:
Hello. I know that very similar issues had been published before but I haven't found any working solution. I don't know if this is very specific to my case but I think it might not be. 

**Issue Descriptions:**

Media control buttons (Play/Pause, Next, Previous) on Logitech K780 using Logitech Options software don't work as expected. If you activate _Show track information in Windows 10_ volume overlay all buttons work, the issue is that next and previous keys jump 2 tracks instead of one. If I disable _Show track information in Windows 10_ option, media keys don't work at all. I also have a Logitech MX Anywhere 2S mouse which have keys that you can configure to work as media keys and in this case, even with the option Show track information in Windows 10 disabled they work normally to play/pause and change tracks. GPMDP is getting different signals from each device but if you activate _Show track information in Windows 10_ even the mouse previous and next keys jump 2 tracks instead of one. This is all with the Playback API activated and JSON API disabled. If I left JSON enable media keys work on the mouse but no the keyboard. I don't know if there is already a solution that works with Logitech Options but I can't find it. I found a solution for Logitech SetPoint but is not applicable in this case. 

Thanks.

**OS:**

Windows 10 1903

**GPMDP Version:**

4.6.1

stefanjm:
Using media next/previous buttons on my acer laptop keyboard also causes it to jump over 1 track.

gcfrxbots:
[MediaKeyFix.zip](https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/files/3513592/MediaKeyFix.zip)

I made an AHK file to somewhat fix this. It'll keep the play/pause button from doublepressing (and thus making it useless) and will make the 'previous track' button work, however due to AHK simulating the keypresses I'm unable to get the menu to appear. The skip track will still double skip, but if you see a song you want then you can go back easily.

Is there any ETA on a fix for this?

schaffrath94:
I have the same issue. Its very annoying... Any sign of fixing the problem?
I use the Corsair K70.

eVen-gits:
Just to add to this:

It's also not working in win8.1 or on linux (kubuntu 18.04 LTS).

RubixSolver99:
I am having the same issue, I cannot use the media keys on my keyboard or my programmed macros on my mouse, without double-pressing the skip/prev/play/pause buttons.

[...]

PR #3748, opened 2019-12-23
Fix Windows 10 Overlay Next/Prev. Media Key Duplication Issue

### LINKED ISSUE #2364: Make W10 Media Service actually usable

### LINKED ISSUE #3293: [4.6.1] Windows Multimedia Keys

Kytech:
This code change will address the double-skipping bug with the Windows 10 media overlay enabled that is mentioned in several issues.

When the Windows 10 media overlay integration is used, Windows will call the event handler function registered for the media overlay's next/previous buttons whenever any physical hardware next/prev buttons are pressed. I added code to check if overlay integration is turned on. If overlay is turned on, the program will not register a keypress handler using electron's KeyRegisterFn to prevent electron from calling the program's next/prev function in addition to Windows firing it's call to the next/prev function when a media key is pressed. We want only Windows to handle the keypress event with the overlay turned on. If both handle the event, the result is the double-skip bug.

This change resolves the following issues:
Closes #2364 
Closes #3293

Addresses part of the following issue:
Issue #3598
This issue also mentions the double-skipping bug, but also describes an issue with the play/pause functionality, which is similar to this problem, though it is being caused in a different part of the program.


jostrander:
This would indeed fix the double skipping problem on initial startup, but what about if the user unchecks the Windows 10 media button in settings? It would leave them without keyboard shortcuts, right?

Kytech:
@jostrander If the user unchecks the Windows 10 media buttons in settings, the program will register shortcuts using Electron's key press handler instead of using the overlay to handle them. The if statement on line 15 that wraps the electron key press event handler registration checks to see if the program settings are configured to use the Win10 media keys and if the program is running on Windows 10 specifically. If the Win10 media service is disabled or the user is not on Windows 10, they will still have keyboard shortcuts without the Win10 media keys since it will then make Electron handle the media keys. I ran a test on this myself and verified this. You could also test this if you'd like with the AppVeyor build for this PR to confirm that it is working as intended.

jostrander:
I'm not using windows at the moment, but I believe you have to restart GPMDP after selecting the media service, correct? If that's the case then the code makes sense. 

I was concerned that an 'on-the-fly' change in settings wouldn't be registered because I saw no event listener for that change being modified. However I believe now that the setting is changed, saved, then applied on restart.

Kytech:
Yes, that is correct. The current behavior for GPMDP's media key settings is that it requires a restart. Since that is the current behavior, I kept it that way. The settings change is applied on restart, as indicated by the settings UI.

Kytech:
I've noticed that the play/pause button has some general issues dispatching events, though that seems to be from a different cause that I plan to investigate a bit myself in another pull request. I originally designed this fix in an older version of GPMDP and it worked perfectly, but after I pulled in the newer code I noticed the play/pause issue. I'm thinking of opening another pull request for that after I have a chance to investigate that issue as well. The skip next/forward seems to be consistent though. Wouldn't inconsistencies with the media service sending events or not be a separate issue? I'm thinking this pull request is to just fix part of the issue since it seems like a bit more of a larger problem.On Dec 27, 2019 7:36 PM, Samuel Attard <notifications@github.com> wrote:@MarshallOfSound requested changes on this pull request.

I've experimented with this before and on some machines (one of mine in particular) the media service does not always emit events when some applications are in focus (games in particular).  I'd want to test this extensively again before landing

xCSxXenon:
Is this just waiting for the play/pause issue to be fixed as well? Really excited to get a fix for the double skip bug!

Kytech:
@MarshallOfSound I've just updated the code base of my PR to be inline with the 4.7.1 release. After testing once I merged in the updates, The play/pause issue no longer occurs. Track movement and media keys work perfectly, as does the system media transport controls integration when they are fully enabled.

However, I noticed that when I disable the "Show Media Information in System Overlay" option, or disable the system media transport controls option entirely, media keys don't work at all. It sounds like the alternate/original method used to respond to media keys is not re-enabling with the changes I made. This issue sounds a lot like #3598, though I may have been able to make everything behave more consistently. I suspect the issue here is due to the change in the settings menu from when I originally made this fix, so I probably need to adjust how this fix works with the settings options.

Also, when this is resolved, it will also close #2687 since the system media transport controls can facilitate this.

One final note, When I last built this project on my dev machine, I was receiving warnings that the current implementation we are using for integrated system media transport controls (The MediaControl class) is deprecated in favor of using the SystemMediaTransportControls class and that the MediaControl class is not guaranteed to work on versions of Windows beyond 8.1. This may contribute to the reason why it works in many cases, yet doesn't in a fair amount of cases as well. Best option is probably to get this working as a toggle-able option for now and then update the implementation to use SystemMediaTransportControls in a separate pull request.

Kytech:
@MarshallOfSound, I've just updated my patch to work with the current settings menu configuration. I also noticed that Windows does see a media playback command and a media key press as different events. I noted this when I configured my laptop trackpad (uses the precision drivers and can configure gestures through the Windows settings) to use a three finger side swipe for media playback control options next track and previous track. When I enabled system media transport controls, this worked as expected. However, without them, they did not function. The keybindings as implemented when system media transport controls are off is to look for a specific keyboard keypress of a next or previous key. The default behavior of Windows's media keys when a program is not binding to them is to essentially trigger a media playback command. I noticed the same thing with Windows Media Player when trying to use my trackpad configuration, which was coded before System Media Transport Controls was built as an API for Windows.

I'm thinking that this may be why the system media transport controls may not always work in certain games. They may register hooks to the keypress events, which then prevents that keypress from being forwarded to Windows to generate a system media transport controls event. I could especially see this being the case where the media playback keys are implemented as alternate functions of the Function/F keys on some machines/keyboards, depending on how they implement this behavior and how some games deal with it. I'm planning to test my patch with something like Auto-Hotkey to see how it behaves. It seems like some games likely interfere with anything using the new system media controls, so making this option controllable is still our best option.

I think my main goal for this PR is to get an implementation that is in line with the current "Windows 10" way of doing things so that we're able to handle media playback controls in a way that plays best with the OS without the annoying double-skipping bug that we were experiencing before this small tweak in how media keys are handled with the System Media Controls Setting turned on.

Kytech:
Another thought as well, @MarshallOfSound, in the settings menu, the "Show Track information in Windows 10 Volume Overlay" settings seems to be redundant since the "Enable the Windows 10 System Media Service" setting essentially has the same effect. The only difference I see is the latter option states that it will only show playback info on the lock screen. However, Windows does not provide a way to do this without also displaying track info in the volume overlay. Should this setting option be removed or is it's intended purpose to do something else? I noted that disabling track info in the volume overlay has the same effect as disabling the Windows System Media Service entirely.

Kytech:
@MarshallOfSound does the issue with the Windows System Media Transport Controls also occur with programs such as Spotify? I know that Spotify uses a similar integration as the one we have with the Windows System Media Transport Controls, so if the same issue occurs with Spotify and those games you are having issues with GPMDP and the media transport controls, I think we could conclude that the issue is those games instead of our implementation. I also made a few tweaks to the way this fix interfaces with the settings since I found a few unexpected behaviors before I made that change. Perhaps this might help with some of the issues we are experiencing?

Kytech:
@MarshallOfSound I do note that you stated in the first issue that I tagged in the original PR that the thought of making the overlay optional, which has been done, would allow us to work around poorly designed programs that interfere with the Windows System Media Transport Controls. I know that Spotify also makes that integration optional as well, so I think we're on the right track concerning having this optional. However, I've noticed from my use of newer hardware that many of the newer hardware's media control features actually dispatch media control actions through the system media transport controls instead of emulating a hardware keypress.

To provide an example, newer windows PCs that support precision trackpads allow for advancing through tracks using gestures on the trackpad. The only way I can get that to work on my surface book 2 (which uses precision trackpad drivers) with GPMDP is to enable the media transport controls. My logitech mouse behaves similarly when I configure it's macros using Logitech's options software. I'm thinking that we should get this fix pushed through as it will appear that media transport controls are entirely broken on newer hardware that dispatches media transport controls through the system media transport controls if it is not fixed. Getting this fix applied should be the most reliable way to have clean media key integration in circumstances for the vast majority of Windows 10 users while the option of switching it off and using global listeners will take care of the other instances where a game may interfere with the functionality. Seeing how Microsoft is pushing for the system media transport controls as the new standard, I think it would be good to incorporate this so that GPMDP plays nice with newer hardware. It seems like we've found a fix that will still have the flexibility to work with all situations by making this a toggle-able option. 

I would really like to see this incorporated into the program as I feel like this program has been amazing for my music listening needs and I know a lot of others feel the same way and have experienced the same or similar issues as myself. I know you're also wanting to get a final fix for this issue as well. I too have also seen a few occasional cases where I've had to switch the setting off and on for the system media transport controls, but I get best results that work about 99% of the time for me by using the media transport controls with this fix. I actually get more reliable results on my surface book 2 and even my desktop with the overlay enabled than with it disabled, but the option to switch it off has served as a reliable fix for those cases where I have to work around other programs.

If there are any other cases you would like me to test, I'd be happy to take a look at them. Please let me know what else I can do to get this moved along.

PR #3762, opened 2019-12-28
Fix translation updating

reduckted:
There's a bug in `vendor/update-translations.js` that results in it running out of memory due to the current state of some of the translation files. I'm not 100% sure, but I think it might be caused by some property keys being out of order in some of the locale files.

I've re-written this script to use the [`diff`](https://www.npmjs.com/package/diff) module, which makes it a lot simpler.

The logic is:
1. Read the `en-US.json` file and change all property values to an empty string.
2. Read the locale file being updated and keep the original lines.
3. Make a copy of those lines with all property values set to an empty string.
4. Compute the diff between the lines (without property values) from `en-US.json` and the lines (without property values) from the locale file.
5. For each diff segment:
    - If the segment has not changed, take the original lines from the locale file for that range of the diff.
    - If the segment has been "added", then it means the lines in the locale file should not be there, and they can be skipped over.
    - If the segment has been "removed", then it means the lines from `en-US.json` need to be added to the locale file.
6. After all diff segments have been processed, write the resulting set of lines back to the file.

I've also run the script to get all of the locale files back in sync.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,7 @@
     "coveralls": "^2.11.12",
     "cross-env": "^2.0.1",
     "devtron": "^1.0.1",
+    "diff": "^4.0.1",
     "electron": "3.1.8",
     "electron-devtools-installer": "^2.2.4",
     "electron-mocha": "^3.0.0",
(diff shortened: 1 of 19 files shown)


reduckted:
I'm closing this since I don't use GPM or YouTube Music anymore, but if anyone is still interested in this, feel free to reopen.

PR #3763, opened 2019-12-28
Add error message and reload button when initial page fails to load

### LINKED ISSUE #2769: What should happen to the windows application when there is no active internet connection!!
**OS:**

**Issue Descriptions:**

**Steps to Reproduce:**

reduckted:
Fixes #2769.

At startup, if the initial page of Google Play Music fails to load, instead of just showing a blank page, a "loading error" page is shown. This page has a button that allows you to reload the main window, causing the application to attempt to load Google Play Music again. The page will also automatically reload the main window after 15 seconds.

image

### DIFF
--- a/src/_locales/cs.json
+++ b/src/_locales/cs.json
@@ -76,6 +76,11 @@
   "playback-label-im-feeling-lucky": "",
   "playback-label-toggle-repeat": "",
 
+  "player-load-failed-gpm": "",
+  "player-load-failed-ytm": "",
+  "player-load-failed-retry": "",
+  "player-load-failed-countdown": "",
+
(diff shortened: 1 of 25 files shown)


reduckted:
What's the usual procedure with the translations? As you can see, I've added blank values to each of the locale files. Should I use Google Translate to add the translations, or do we just leave them blank and wait for someone else to fill them in through another PR?

MarshallOfSound:
For translation strings just leave them blank and let other folks fill them in if they want to.

Looks like a good feature, I'll take a look at the code soon

reduckted:
I'm closing this since I don't use GPM or YouTube Music anymore, but if anyone is still interested in this, feel free to reopen.

PR #3775, opened 2020-01-10
Feature: Slack status for multiple workspaces

SpenserJ:
This reworks the Slack status integration to support multiple tokens/workspaces, and adds a new settings field `TextListFieldSettings` that uses MaterialUI to render an array of strings as individual entries, with a text field for adding new values at the bottom. Adding a new token will update the status for all Slack clients, while removing a token will only reset the status for that token.
2020-01-09T18:09:41-07:00
2020-01-09T18:10:08-07:00

### DIFF
--- a/src/_locales/en-US.json
+++ b/src/_locales/en-US.json
@@ -135,7 +135,7 @@
   "settings-options-style-dialog-css-files": "CSS Files",
   "settings-options-style-dialog-all-files": "All Files",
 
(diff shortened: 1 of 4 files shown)


jostrander:
Is this something we should still pursue? Slack Legacy tokens will no longer be available after May 5, 2020 according to the generation page. 

https://api.slack.com/legacy/custom-integrations/legacy-tokens

PR #3824, opened 2020-04-22
Add keyboard shortcut for lyrics

### LINKED ISSUE #3818: Easy enhancement: Add a configurable shortcut for opening lyrics ✨
Hey! It would be cool if we could edit the shortcut for the lyrics, just like we can edit like, play/pause etc.

jackyboygood:
Closes #3818


jostrander:
Obviously this is configurable, but should we have both the menu shortcut and this?

https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/blob/master/src/main/features/core/applicationMenu.js#L80

MarshallOfSound (approved):

PR #3825, opened 2020-04-22
Toggle YTM button in settings

### LINKED ISSUE #3765: Option to hide the 'Switch to YTM' button
Please add an option to hide the Switch to YTM button, or to at least to customise it so it can be made dark, thank you.

jackyboygood:
Closes #3765 
How should I generate the language text? Currently I have it hardcoded as English but I can add the translations

### DIFF
--- a/src/_locales/cs.json
+++ b/src/_locales/cs.json
@@ -126,6 +126,8 @@
   "settings-option-skip-bad-songs": "",
   "settings-option-try-supported-channel-layouts": "Experimentální podpora více audio kanálů, jako je 5.1",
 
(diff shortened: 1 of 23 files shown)


jostrander:
As far as I know, you add the key for every language, but leave the value blank for everything but english. It will show the english translation by default and translators can submit their translations from there.

MarshallOfSound (approved):

Alipoodle:
Isn't this fix not needed now considering that the button should be removed due to the end of GPM?

jostrander:
Technically correct, yes.

PR #3840, opened 2020-05-16
Feature/bulgarian lanuage support

RosenDev:


RosenDev:
Not sure about Bulgarian code in provider.js ('bg').

